PR #17, opened 2019-01-25
Fix dismiss recommendation operation

javoeria:
Hello, i'm trying to use the operation of dismiss a recommendation but it raises an error:

`ArgumentError: No operation found with name DismissRecommendation`

### DIFF
--- a/lib/google/ads/google_ads/v0/services/recommendation_service_client.rb
+++ b/lib/google/ads/google_ads/v0/services/recommendation_service_client.rb
@@ -286,11 +286,11 @@ module Google
             #   operations will return errors. If false, operations will be carried in a
             #   single transaction if and only if they are all valid.
             #   Default is false.
-            # @param operations [Array<Google::Ads::GoogleAds::V0::Services::DismissRecommendationRequest::DismissRecommendationOperation | Hash>]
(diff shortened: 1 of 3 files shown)


lib/google/ads/google_ads/v0/services/recommendation_service_client.rb
```diff
@@ -286,11 +286,11 @@ def apply_recommendation \
             #   operations will return errors. If false, operations will be carried in a
             #   single transaction if and only if they are all valid.
             #   Default is false.
-            # @param operations [Array<Google::Ads::GoogleAds::V0::Services::DismissRecommendationRequest::DismissRecommendationOperation | Hash>]
+            # @param operations [Array<Google::Ads::GoogleAds::V0::Services::DismissRecommendationOperation | Hash>]
```
mcloonan: Please don't change anything in the v0/ folder. These are all automatically-generated.

javoeria:
I don't think it can be fixed without those changes in v0 folder but thanks for the quick response.

lib/google/ads/google_ads/v0/services/recommendation_service_client.rb
javoeria: I don't think it can be fixed without those changes in v0 folder but thanks for the quick response.

Issue #22, opened 2019-02-01
NPE in https://github.com/googleads/google-ads-ruby/blob/master/lib/google/ads/google_ads/logging_interceptor.rb


dmitchell:
`most_recent_error.status_details` can be nil. Either change `Google::Gax::Grpc.deserialize_error_status_details` to return an empty list or change L55 to `most_recent_error.status_details&.each`

PR #24, opened 2019-02-01
NPE for status_details

dmitchell:
https://github.com/googleads/google-ads-ruby/issues/22

### DIFF
--- a/lib/google/ads/google_ads/logging_interceptor.rb
+++ b/lib/google/ads/google_ads/logging_interceptor.rb
@@ -52,7 +52,7 @@ module Google
             response_message = "Incoming response (errors): \n"
 
             most_recent_error = Google::Gax::GaxError.new('')
-            most_recent_error.status_details.each do |detail|
+            most_recent_error.status_details && most_recent_error.status_details.each do |detail|
               if detail.is_a?(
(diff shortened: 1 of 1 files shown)


lib/google/ads/google_ads/logging_interceptor.rb
```diff
@@ -52,7 +52,7 @@ def request_response(request:, call:, method:, metadata: {})
             response_message = "Incoming response (errors): \n"
 
             most_recent_error = Google::Gax::GaxError.new('')
-            most_recent_error.status_details.each do |detail|
+            most_recent_error.status_details&.each do |detail|
```
mcloonan: The &. operator was added in Ruby 2.3. We'd prefer not to restrict support to earlier versions unnecessarily, even if they are rarely used. Can you implement this a different way?

dmitchell:
Done but github won't let me comment on the PR

mcloonan:
Looks like a lot more changes made it in this time somehow. Can you limit it to just the NPE fix? Thanks!

dmitchell:
Yeah, my ide corrected interpolated constant strings to non-interpolated
.... Reverted

Issue #23, opened 2019-02-01
google_ads_client.rb must cast login_customer_id to string


dmitchell:
`            headers[:"login-customer-id"] = login_customer_id` You must add `.to_s` otherwise it gives an error downstream about headers must be strings

And, you may as well cast it to int if it was passed as string in order to do the bounds check

PR #25, opened 2019-02-01
Header values must be strings

dmitchell:
https://github.com/googleads/google-ads-ruby/issues/23


mcloonan:
Thanks for fixing this!

PR #29, opened 2019-02-04
Fixing logging interceptor for customer service requests.

### LINKED ISSUE #26: Get customer request not working in v0.7
After the v0.7 update, the method "get_customer" of customer service is not working and it gets an error:

`undefined method 'customer_id' for #<Google::Ads::GoogleAds::V0::Services::GetCustomerRequest>`

You can check it in the "get_account_information.rb" example. Thanks.

mcloonan:
Fixes #26

### DIFF
--- a/lib/google/ads/google_ads/logging_interceptor.rb
+++ b/lib/google/ads/google_ads/logging_interceptor.rb
@@ -29,10 +29,12 @@ module Google
         end
 
         def request_response(request:, call:, method:, metadata: {})
+          customer_id = "N/A"
           customer_id = request.customer_id if request.respond_to?(:customer_id)
           # CustomerService get requests have a different format.
-          customer_id =
-              request.resource_name.split('/').last if customer_id.nil?
+          if request.respond_to?(:resource_name)
+            customer_id = request.resource_name.split('/').last
+          end
           summary_message =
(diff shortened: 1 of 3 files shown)


voke:
I merged this locally and it looks like this will break the `list_accessible_customers` method.

**Error:**
```
gems/google-ads-ruby-26d6368187f6/lib/google/ads/google_ads/logging_interceptor.rb:35 - undefined method `resource_name' for #<Google::Ads::GoogleAds::V0::Services::ListAccessibleCustomersRequest:0x00007fbe95718ff8>
```

mcloonan:
Thanks for pointing this out; I'll make it more robust before merging.

lib/google/ads/google_ads/logging_interceptor.rb
```diff
@@ -29,9 +29,15 @@ def initialize(logger)
         end
 
         def request_response(request:, call:, method:, metadata: {})
+          customer_id = "N/A"
```
wihl, line 32: I was reviewing the Python version of this (https://github.com/googleads/google-ads-python/pull/262#discussion_r407575545). Are you sure that "N/A" is the correct default response given that '/' is the delimiter between resource name parts? Perhaps "NA" or "None"?

PR #32, opened 2019-02-07
only add LoggingInterceptor if logger exists

### LINKED ISSUE #31: LoggingInterceptor crashes if logger config is not provided
When the ads client is initialized it's possible to have the default logger fail to be created. The client will return but will fail on subsequent calls in which the LoggingInterceptor is invoked. I would expect the LoggingInterceptor to just never be used if no logger is set

https://github.com/googleads/google-ads-ruby/blob/50e510c81da896b1569906688e583a5e30a19d10/lib/google/ads/google_ads/google_ads_client.rb#L78-L83

https://github.com/googleads/google-ads-ruby/blob/50e510c81da896b1569906688e583a5e30a19d10/lib/google/ads/google_ads/google_ads_client.rb#L99-L109

jbodah:
Fixes #31. Please let me know if there is an example test I can use to add a test for this

### DIFF
--- a/lib/google/ads/google_ads/google_ads_client.rb
+++ b/lib/google/ads/google_ads/google_ads_client.rb
@@ -125,9 +125,12 @@ module Google
             unless service_path.nil? || service_path.empty?
               const_set('SERVICE_ADDRESS', service_path.freeze)
             end
-            logging_interceptor =
-                Google::Ads::GoogleAds::LoggingInterceptor.new(logger)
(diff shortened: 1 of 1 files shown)


jbodah:
I signed it!

mcloonan:
I would like to get this in but there's still a problem with the CLA. If you get a chance, please try to follow the instructions it provided again so we can merge this branch!

jbodah:
@mcloonan should be all set now - the CLA page was buggy when it comes to having multiple accounts signed in

Issue #40, opened 2019-03-06
Feature Request: provide with intermediate objects to avoid client global


apneadiving:
For now, pretty much everything is attached to the client for "convenience" sake.

Yet as a developer, I would not like to need a client object to build the operations: in the end its more a global value than a convenient one.

Basically I would like to be able to create an operation, use an enum, use wrappers without the client.

For instance:
```
# in addition of 
client = ::Google::Ads::GoogleAds::GoogleAdsClient.new
client.operation(name)
# I'd like to be able to use
::Google::Ads::GoogleAds::GoogleAdsClient.operation(name, version)
# or maybe something like
proto_lookup = ::Google::Ads::GoogleAds::ProtoLookup.new(version)
proto_lookup.operation(name)
```
Indeed, creating an operation should not rely on some api configuration triggered during client instantiation.

One way to achieve this would be to have `proto_lookup_utils` at the class level. 

Another way would be to move lookup utils in an independent object.

Would you accept either move?

apneadiving:
the `ProtoLookup` would be the easiest way to proceed, I could suggest a pull request if I have a green

mcloonan:
Thanks for implementing this!

PR #42, opened 2019-03-08
move lookup_util to its own class

apneadiving:
Explanations in #40

### DIFF
--- a/lib/google/ads/google_ads/google_ads_client.rb
+++ b/lib/google/ads/google_ads/google_ads_client.rb
@@ -73,7 +73,7 @@ module Google
 
         DEFAULT_CONFIG_FILENAME = 'google_ads_config.rb'
 
-        attr_reader :logger
+        attr_accessor :logger
         attr_writer :lookup_util
 
         def initialize(config_path = nil, &block)
@@ -194,12 +194,6 @@ module Google
         # Returns a reference to the PathLookupUtil to generate resource names.
         def path(version = default_api_version)
           lookup_util.path(version)
-        end
-
-        # Set the logger to use. This will only take effect on services fetched
-        # after setting this value.
-        def logger=(logger)
-          @logger = logger
         end
 
         private
--- a/lib/google/ads/google_ads/google_ads_client.rb
+++ b/lib/google/ads/google_ads/google_ads_client.rb
@@ -70,7 +70,6 @@ module Google
   module Ads
     module GoogleAds
       class GoogleAdsClient
-        DEFAULT_API_VERSION = :V1
 
         DEFAULT_CONFIG_FILENAME = 'google_ads_config.rb'
 
@@ -121,7 +120,7 @@ module Google
         # :Campaign will return an instantiated CampaignServiceClient.
         #
         # Raises ArgumentError if no service can be found for the provided type.
-        def service(name, version = DEFAULT_API_VERSION)
+        def service(name, version = default_api_version)
           service_path = ENV['GOOGLEADS_SERVICE_PATH']
 
           # We need a local reference to refer to from within the class block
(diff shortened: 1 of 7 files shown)


apneadiving:
I signed it!

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -73,7 +73,7 @@ class GoogleAdsClient
 
         DEFAULT_CONFIG_FILENAME = 'google_ads_config.rb'
 
-        attr_reader :logger
+        attr_accessor :logger
```
mcloonan: Can you please undo the changes for attr_reader vs. attr_accessor with the logger? I specifically separated them out partially for the comment to have a place to exist but also because there's a chance that more needs to be done on setting the logger in the future than simple assignment.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -69,12 +70,11 @@ module Google
...
     module GoogleAds
       class GoogleAdsClient
-        DEFAULT_API_VERSION = :V1
-        KNOWN_API_VERSIONS = [:V0, :V1]
 
         DEFAULT_CONFIG_FILENAME = 'google_ads_config.rb'
 
-        attr_reader :logger
+        attr_accessor :logger
+        attr_writer :lookup_util
```
mcloonan: I'm not sure why this has to exist.

lib/google/ads/google_ads/lookup_util.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+
+        # Returns a reference to the PathLookupUtil to generate resource names.
+        def path(version)
+          path_lookup_util(version)
+        end
+
+        private
+
+        attr_reader :proto_lookup_utils, :path_lookup_utils
```
mcloonan: I don't think these need to be public.

apneadiving: yes they are private indeed :)

apneadiving:
my point is the following: creating operations should have nothing to do with the client. In other words: I should not have to worry about credentials to create operations etc. 

I understand it is handy to have methods on client, so lets keep them but it is just bonus: in OOP one object should be responsible for one thing.

lib/google/ads/google_ads/google_ads_client.rb
apneadiving: ok sure

lib/google/ads/google_ads/google_ads_client.rb
apneadiving: the reason is the following: I intend to create a lookup object to create operations.

Then later, I intend to create the client.

Because I dont want to lose all the memoization done by the lookup object, I'd like to pass it to the client.

apneadiving:
I completely get the service particular case. It makes sense to keep part of the code there as it currently is.

apneadiving:
We could rename service to raw_service ine the lookup class. Would avoid confusion

lib/google/ads/google_ads/google_ads_client.rb
mcloonan: Can you just add a comment here explaining that, and then the rest of this PR looks good. Sorry the review is taking so long; this is the first substantial PR from a community member :)

apneadiving: done! I totally get some time is necessary to include contributions :)

mcloonan: Can you clean up to:

    # Allow setting the lookup_util manually for users who use it before creating the client.

Thanks!

apneadiving: done ;) bonus point: it allows caching the lookups but I'd rather not mention this.

PR #43, opened 2019-03-08
Add an example for generating a report to CSV.

samphippen:


examples/reporting/campaign_report_to_csv.rb
```diff
@@ -0,0 +1,136 @@
...
+#
+# This code example illustrates how to get metrics about a campaign and 
+# serialize the result as a CSV file.
+
+require 'optparse'
+require 'time'
+require 'csv'
+require 'google/ads/google_ads'
+
+PAGE_SIZE = 50
```
AnashOommen: Please make this pagesize considerably bigger, since we are dealing with reports which usually will have lots of rows.

examples/reporting/campaign_report_to_csv.rb
```diff
@@ -0,0 +1,136 @@
...
+      ORDER BY segments.date DESC
+  QUERY
+
+  response = ga_service.search(customer_id, query, page_size: PAGE_SIZE)
+  # convert the Google Ads response rows in to CSV ready hash objects
+  csv_rows = response.map { |row| result_row_as_csv_hash(row) }
+
+  CSV.open(target_filepath, "wb") do |csv|
+    # use the keys of the first csv_row as a header row
+    csv << csv_rows.first.keys
```
mcloonan: I think we may want to output the entire path to the field (i.e. what you specify in the select clause) rather than just the keys you use in the hash.

examples/reporting/campaign_report_to_csv.rb
```diff
@@ -0,0 +1,136 @@
...
+    opts.separator 'Help:'
+
+    opts.on_tail('-h', '--help', 'Show this message') do
+      puts opts
+      exit
+    end
+  end.parse!
+
+  begin
+    # write CSV output to campaign_report_to_csv.csv
```
mcloonan: I think it would be nice to allow the user to specify a filepath on the command line as well.

examples/reporting/campaign_report_to_csv.rb
```diff
@@ -0,0 +1,136 @@
...
+    output_file_path = __FILE__.gsub(".rb", ".csv")
+
+    # the GoogleAdsClient only accepts customer IDs without `-` characters,
+    # so this removes them if the caller of this script copy pasted a customer
+    # id directly from the user interface
+    sanitized_customer_id = options.fetch(:customer_id).tr("-", "")
```
mcloonan: This is pretty nice; at some point can you modify other examples to do this as well?

samphippen: Todo in follow up pull request.

PR #45, opened 2019-03-13
Create an "Add Customer Negative Criteria" example

samphippen:
ported from [this soap example](https://github.com/googleads/google-api-ads-ruby/blob/master/adwords_api/examples/v201809/targeting/add_customer_negative_criteria.rb)

### DIFF
--- a/examples/targeting/add_customer_negative_criteria.rb
+++ b/examples/targeting/add_customer_negative_criteria.rb
@@ -1,0 +1,106 @@
+#!/usr/bin/env ruby
+# Encoding: utf-8
+#
+# Copyright:: Copyright 2017, Google Inc. All Rights Reserved.
+#
+# License:: Licensed under the Apache License, Version 2.0 (the "License");
+#           you may not use this file except in compliance with the License.
+#           You may obtain a copy of the License at
+#
+#           http://www.apache.org/licenses/LICENSE-2.0
+#
+#           Unless required by applicable law or agreed to in writing, software
+#           distributed under the License is distributed on an "AS IS" BASIS,
+#           WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+#           implied.
+#           See the License for the specific language governing permissions and
+#           limitations under the License.
+#
+# This example adds various types of negative criteria to a customer. These
+# criteria will be applied to all campaigns for the customer.
+
+require 'optparse'
+require 'google/ads/google_ads'
+
+def add_customer_negative_criteria(customer_id)
+  # GoogleAdsClient will read a config file from
+  # ENV['HOME']/google_ads_config.rb when called without parameters
+  client = Google::Ads::GoogleAds::GoogleAdsClient.new
+
+  tragedy_criterion = client.resource(:CustomerNegativeCriterion)
+  tragedy_criterion.content_label = client.resource(:ContentLabelInfo)
+  tragedy_criterion.content_label.type = :TRAGEDY
+
+  placement_criterion = client.resource(:CustomerNegativeCriterion)
+  placement_criterion.placement = client.resource(:PlacementInfo)
+  placement_criterion.placement.url = client.wrapper.string("http://example.com")
+
+  ops = [
+    {create: tragedy_criterion},
+    {create: placement_criterion},
+  ]
+
+  customer_negative_criterion_service = client.service(
+    :CustomerNegativeCriterion
+  )
+
+  response = customer_negative_criterion_service.mutate_customer_negative_criteria(customer_id, ops)
+
+  # Display the results.
+  response.results.each do |negative_criterion|
+    puts "Customer negative criterion with criterion resource name #{negative_criterion.resource_name} was added"
+  end
+end
+
+if __FILE__ == $0
+  options = {}
+  # The following parameter(s) should be provided to run the example. You can
+  # either specify these by changing the INSERT_XXX_ID_HERE values below, or on
+  # the command line.
+  #
+  # Parameters passed on the command line will override any parameters set in
+  # code.
+  #
+  # Running the example with -h will print the command line usage.
+  options[:customer_id] = 'INSERT_CUSTOMER_ID_HERE'
+
+  OptionParser.new do |opts|
+    opts.banner = sprintf('Usage: add_customer_negative_criteria.rb [options]')
+
+    opts.separator ''
+    opts.separator 'Options:'
+
+    opts.on('-C', '--customer-id CUSTOMER-ID', String, 'Customer ID') do |v|
+      options[:customer_id] = v
+    end
+
+    opts.separator ''
+    opts.separator 'Help:'
+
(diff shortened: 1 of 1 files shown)


examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+#           Unless required by applicable law or agreed to in writing, software
+#           distributed under the License is distributed on an "AS IS" BASIS,
+#           WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+#           implied.
+#           See the License for the specific language governing permissions and
+#           limitations under the License.
+#
+# This example adds various types of negative criteria to a customer. These
+# criteria will be applied to all campaigns for the customer.
+#
```
mcloonan: Remove.

examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+require 'optparse'
+require 'google/ads/google_ads'
+
+def add_customer_negative_criteria(customer_id)
+  # GoogleAdsClient will read a config file from
+  # ENV['HOME']/google_ads_config.rb when called without parameters
+  client = Google::Ads::GoogleAds::GoogleAdsClient.new
+
+  tradgedy_criterion = client.resource(:CustomerNegativeCriterion)
```
mcloonan: Fix the spelling of this variable.

examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+    {create: placement_criterion},
+  ]
+
+  customer_negative_criterion_service = client.service(
+    :CustomerNegativeCriterion
+  )
+
+  response = customer_negative_criterion_service.mutate_customer_negative_criteria(customer_id, ops)
+
+  if response.partial_failure_error != nil
```
mcloonan: unless response.partial_failure_error.nil?

examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+    raise "Got a partial failure #{response.partial_failure_error.inspect} while creating criterea"
```
mcloonan: criterea -> criteria

examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+#
+#           Unless required by applicable law or agreed to in writing, software
+#           distributed under the License is distributed on an "AS IS" BASIS,
+#           WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+#           implied.
+#           See the License for the specific language governing permissions and
+#           limitations under the License.
+#
+# This example adds various types of negative criteria to a customer. These
+# criteria will be applied to all campaigns for the customer.
```
mcloonan: I think the second sentence here is a little confusing. It sounds like you'll be manually applying criteria to each campaign, rather than that they will automatically apply to each campaign since they're set at the customer level. Can we make sure it's clear what exactly is going to happen?

examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+    opts.separator 'Help:'
+
+    opts.on_tail('-h', '--help', 'Show this message') do
+      puts opts
+      exit
+    end
+  end.parse!
+
+  begin
+    add_customer_negative_criteria(options.fetch(:customer_id))
```
mcloonan: The other examples use options[:customer_id]; is there a reason you used fetch instead?

examples/targeting/add_customer_negative_criteria.rb
```diff
@@ -0,0 +1,111 @@
...
+  # the command line.
+  #
+  # Parameters passed on the command line will override any parameters set in
+  # code.
+  #
+  # Running the example with -h will print the command line usage.
+  options[:customer_id] = 'INSERT_CUSTOMER_ID_HERE'
+
+  OptionParser.new do |opts|
+    opts.banner = sprintf('Usage: add_campaigns.rb [options]')
```
mcloonan: This isn't add_campaign.rb. The other examples grab the filename here, e.g.: https://github.com/googleads/google-ads-ruby/blob/master/examples/basic_operations/remove_ad_group.rb#L53

examples/targeting/add_customer_negative_criteria.rb
mcloonan: Although I don't think you can get a partial failure response unless you explicitly set partial failure to true in the request, which you don't do here.

examples/targeting/add_customer_negative_criteria.rb
samphippen: 👍

examples/targeting/add_customer_negative_criteria.rb
samphippen: 👍

examples/targeting/add_customer_negative_criteria.rb
samphippen: I'll remove

examples/targeting/add_customer_negative_criteria.rb
samphippen: 👍

examples/targeting/add_customer_negative_criteria.rb
samphippen: this was copied verbatim from the soap example: https://github.com/googleads/google-api-ads-ruby/blob/master/adwords_api/examples/v201809/targeting/add_customer_negative_criteria.rb#L19. Do you have copy you'd prefer?

examples/targeting/add_customer_negative_criteria.rb
samphippen: `fetch` is generally preferable because it raises if a value is not present: https://ruby-doc.org/core-2.2.0/Hash.html#method-i-fetch. It's generally considered best practice to use `fetch` in all places.

examples/targeting/add_customer_negative_criteria.rb
samphippen: 👍

examples/targeting/add_customer_negative_criteria.rb
samphippen: also, re: unless, I prefer to only use it in the single line form: `return x unless y`. If doing a multi line form, I almost always do `if`, even if the conditional is negative.

examples/targeting/add_customer_negative_criteria.rb
mcloonan: Can you add this (and any other style/best-practices things you notice) to that doc you shared with me about potential improvements? Then feel free to resolve this.

samphippen: Done!

PR #49, opened 2019-03-21
Example: Add Gmail Ad

samphippen:


examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+    opts.on_tail('-h', '--help', 'Show this message') do
+      puts opts
+      exit
+    end
+  end.parse!
+
+  begin
+    add_gmail_ad(
+      options.fetch(:customer_id).tr("-", ""),
+      options[:campaign_id],
```
mcloonan: Might as well use options.fetch for all of them now.

examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+# AdvertisingChannelSubType of DISPLAY_GMAIL_AD. To get ad groups, run
+# get_ad_groups.rb.
+
+require 'optparse'
+require 'date'
+require 'open-uri'
+require 'google/ads/google_ads'
+require 'google/ads/google_ads/v1/errors/errors_pb'
+
+def content_type_to_symbol(content_type)
```
mcloonan: Move this method down by the other one (get_image). Is this definitely a safe way to do this for any valid content type? If not we may want to do a more annoying manual lookup so we can raise an error if we get an unsupported content type. Users often actually re-use example code directly.

examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+    marketing_image_content_type
+  )
+
+  media_file_service = client.service(:MediaFile)
+
+  media_file_logo = client.resource(:MediaFile)
+  media_file_logo.type = :IMAGE
+  media_file_logo.image = client.resource(:MediaImage)
+  media_file_logo.image.data = client.wrapper.bytes(logo_image_bytes)
+  media_file_logo.mime_type = logo_image_content_type.to_sym
```
mcloonan: This is already a symbol.

examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+
+  media_file_logo_op = client.operation(:MediaFile)
+  media_file_logo_op['create'] = media_file_logo
+
+  media_file_marketing = client.resource(:MediaFile)
+  media_file_marketing.type = :IMAGE
+  media_file_marketing.image = client.resource(:MediaImage)
+  media_file_marketing.image.data = client.wrapper.bytes(marketing_image_bytes)
+  media_file_marketing.mime_type = marketing_image_content_type.to_sym
```
mcloonan: Ditto.

examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+  gmail_ad.teaser.logo_image = client.wrapper.string(logo_id)
+
+  gmail_ad.marketing_image = client.wrapper.string(marketing_image_id)
+  gmail_ad.marketing_image_headline = client.wrapper.string("Travel")
+  gmail_ad.marketing_image_description = client.wrapper.string("Take to the skies!")
+
+  ad = client.resource(:Ad)
+  ad.final_urls << client.wrapper.string("http://www.example.com")
+  ad.gmail_ad = gmail_ad
+  ad.name = client.wrapper.string("gmail ad")
```
mcloonan: Add the current time to the name here (same as in other create examples), since names have to be unique within an account. This way the example can be run repeatedly.

examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+  ad_group_ad.ad_group = client.wrapper.string(
+    client.path.ad_group(customer_id, ad_group_id),
+  )
+
+  op = client.operation(:AdGroupAd)
+  op['create'] = ad_group_ad
+
+  ad_group_ad_service = client.service(:AdGroupAd)
+  response = ad_group_ad_service.mutate_ad_group_ads(customer_id, [op])
+  puts "created gmail ad with id #{response.results.first.resource_name}"
```
mcloonan: Use sentence case and punctuation.

examples/advanced_operations/add_gmail_ad.rb
```diff
@@ -0,0 +1,182 @@
...
+end
+
+def get_image(url)
+  open(url) { |f|
```
mcloonan: Aren't do/end preferred on multi-line blocks?

examples/advanced_operations/add_gmail_ad.rb
samphippen: this will work for all types that are supported by our protos, The other thing we could do is assert that the HTTP response gives us the response type we're expecting and raise if it doesn't. WDYT?

examples/advanced_operations/add_gmail_ad.rb
samphippen: I'm following the "semantic blocks" convention. This is an infinite yak shave in the ruby community tho. https://github.com/testdouble/standard/blob/master/README.md#standardrb-the-rules

examples/advanced_operations/add_gmail_ad.rb
mcloonan: That sounds good.

devchas:
@samphippen why is campaign ID needed in this example?

PR #52, opened 2019-03-26
Add an example for product category retrieval.

samphippen:


examples/shopping_campaigns/get_product_bidding_category_constant.rb
```diff
@@ -0,0 +1,140 @@
...
+    product_bidding_category = row.product_bidding_category_constant
+    category = {
+      name: product_bidding_category.localized_name.value,
+      id: product_bidding_category.resource_name,
+      children: []
+    }
+
+    all_categories[category.fetch(:id)] = category
+
+    if parent_id = product_bidding_category
```
mcloonan: Should this be `==`?

If not, then please break into two statements for readability, one setting the value and the other testing it.

examples/shopping_campaigns/get_product_bidding_category_constant.rb
```diff
@@ -0,0 +1,140 @@
...
+  EOD
+
+  ga_service = client.service(:GoogleAds)
+  response = ga_service.search(
+      customer_id,
+      query,
+      page_size: PAGE_SIZE,
+  )
+
+  all_categories = Hash.new do |h, k|
```
mcloonan: Can you add a comment explaining that the given block provides a default if the key isn't found?

The goal of these examples is to provide simple, readable code. We need to strike a balance between doing things in a clever, succinct way and doing things in a way that even a relative beginner can quickly comprehend.

PR #54, opened 2019-03-27
Add tests around the logging interceptor

samphippen:
Also fixes a bug where v1 errors were not printing debug info.

### DIFF
--- a/lib/google/ads/google_ads/logging_interceptor.rb
+++ b/lib/google/ads/google_ads/logging_interceptor.rb
@@ -16,6 +16,8 @@
 #
 # Interceptor to log outgoing requests and incoming responses.
 
+require 'google/ads/google_ads/v1/errors/errors_pb'
+require 'google/ads/google_ads/v0/errors/errors_pb'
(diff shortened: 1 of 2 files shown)


lib/google/ads/google_ads/logging_interceptor.rb
```diff
@@ -16,13 +16,20 @@
...
+require 'google/ads/google_ads/v1/errors/errors_pb'
+require 'google/ads/google_ads/v0/errors/errors_pb'
 require 'grpc/generic/interceptors'
 require 'json'
 
 module Google
   module Ads
     module GoogleAds
       class LoggingInterceptor < GRPC::ClientInterceptor
+        INTERESTING_ERROR_CLASSES = [
```
mcloonan: Can we generate this using the KNOWN_VERSIONS array in Google::Ads::GoogleAds instead, so it updates automatically when we modify that canonical list?

samphippen: done!

PR #55, opened 2019-03-27
Refactor logging interceptor

samphippen:
Each refactor is broken in to it's own commit, also fixes a circular require between google_ads.rb and google_ads/logging_interceptor.rb

### DIFF
--- a/lib/google/ads/google_ads/logging_interceptor.rb
+++ b/lib/google/ads/google_ads/logging_interceptor.rb
@@ -30,7 +30,7 @@ module Google
         }
 
         def initialize(logger)
-          # Don't propogate args, parens are necessary
+          # Don't propagate args, parens are necessary
           super()
           @logger = logger
         end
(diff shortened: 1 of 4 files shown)


lib/google/ads/google_ads/logging_interceptor.rb
```diff
@@ -30,24 +30,55 @@ class LoggingInterceptor < GRPC::ClientInterceptor
         }
 
         def initialize(logger)
+          # Don't propogate args, parens are necessary
```
mcloonan: "propagate"

lib/google/ads/google_ads/logging_interceptor.rb
```diff
@@ -30,42 +30,54 @@ class LoggingInterceptor < GRPC::ClientInterceptor
...
+          # the current exception as its cause, and then parses details
+          # out of that exception. So this sets it up so that
+          # most_recent_error.status_details contains useful information about
+          # our failure
+          most_recent_error = Google::Gax::GaxError.new('')
+          response_message = "Incoming response (errors): \n"
+
+          return response_message unless most_recent_error.status_details
+
+          formatted_details = most_recent_error.status_details.select { |detail|
```
mcloonan: We have another bug open because status_details can sometimes be a string. Can we address that as well while we're here?

samphippen: 👍

PR #57, opened 2019-03-27
Make logger fully injectable

### LINKED ISSUE #34: Injectable config.logger
Would it be possible to have an injectable `config.logger` (rather than just the stream/level)? We try to reuse `Rails.logger` or one of our custom logger instances whenever possible

samphippen:
Fixes #34


jbodah:
Looks good to me 👍 🚢

PR #58, opened 2019-03-27
Update logging_interceptor.rb

apneadiving:
use block syntax to compute logs only when meant to be logged


lib/google/ads/google_ads/logging_interceptor.rb
```diff
@@ -39,14 +39,14 @@ def request_response(request:, call:, method:, metadata: {})
...
             response = yield
 
-            @logger.info(build_summary_message(request, call, method, false))
-            @logger.debug(build_request_message(metadata, request))
-            @logger.debug(build_success_response_message(response))
+            @logger.info { build_summary_message(request, call, method, false) }
+            @logger.debug { build_request_message(metadata, request) }
+            @logger.debug { build_success_response_message(response) }
             response
           rescue Exception
```
apneadiving: best practive is to only rescue errors from StandardError, but its another question

samphippen: In this case I'm totally ok with a blanket rescue because we reraise at the end of the call, wdyt @apneadiving?

apneadiving: you are right!

apneadiving:
updated from master

apneadiving:
updated again

PR #70, opened 2019-04-02
factories

samphippen:
The idea here is to replace the ProtoLookupUtil class with something
that uses method calls instead of magic symbol lookup. In addition to
this, it'd be ideal if this was done statically insteaad of dynamically.
The reason for a preference over dynamic code (say method missing or
define method) is that then IDEs and other tooling can provide
autocompletes for users of the library.

So, the solution: it loads in all the protos under google ads (and then
some in the blast raadius) and after each one is defined, it stores
them in an array of known protos. Then it renders out an erb template
with that data.

Right now this is a hacky prototype but I was easily able to port a
campaign addition script over. Next I'll try somethinig meatier and
properly defining all resources/services.


scripts/generate_resources_rb.rb
```diff
@@ -0,0 +1,40 @@
+require 'ripper'
```
mcloonan: I know this is just a PoC, but some of the stuff in the "resources" folder is actually located in "common", and a few things are even in "services". I don't think it's an important distinction to an end-user which one they're from, which is why I chose to mix them all together.

Specifically, programmatically choosing which things from "services" to include and which are irrelevant is likely going to be a challenge, especially since operations are also hosted in "services" and we'd probably want them stored elsewhere.

examples/basic_operations/add_campaigns.rb
```diff
@@ -25,11 +25,12 @@ def add_campaigns(customer_id)
   # GoogleAdsClient will read a config file from
   # ENV['HOME']/google_ads_config.rb when called without parameters
   client = Google::Ads::GoogleAds::GoogleAdsClient.new
+  resource = Google::Ads::GoogleAds.resource
```
mcloonan: Let's call this "resources" and also add a convenience access to it on the client.

lib/google/ads/google_ads/resource.rb
```diff
@@ -0,0 +1,1566 @@
+module Google
+  module Ads
+    module GoogleAds
+      def self.resource
+        @resource ||= Module.new do
+        
+          def self.http
+            require "google/api/http_pb.rb"
```
mcloonan: I'm not exactly sure why this is getting included. Glancing through the protos in the resources folder I don't see any mentions to this specifically (same for other things from google/api in this file). If we do want to include it for some reason maybe it should go in its own module rather than mixed with Google Ads API stuff.

lib/google/ads/google_ads/resource.rb
```diff
@@ -0,0 +1,1566 @@
...
+            Google::Api::HttpRule.new
+          end
+        
+          def self.custom_http_pattern
+            require "google/api/http_pb.rb"
+            Google::Api::CustomHttpPattern.new
+          end
+        
+          def self.double_value
+            require "google/protobuf/wrappers_pb.rb"
```
mcloonan: Having access to these wrapper things will be moot once the new version of the protobuf library is out since the objects will support `*_as_value` methods, so something like this would work: `campaign.name_as_value = "new name"`.

We can still expose them in case people want to explicitly use wrappers but I'd also move these to their own module with the google/api stuff.

examples/factories/various_factories_usage.rb
```diff
@@ -0,0 +1,171 @@
...
+  # updates also work with only a resource name, so let's pull one out and
+  # then update it
+  ga_service = client.service.google_ads
+  res = ga_service.search(customer_id, <<~EOQUERY)
+  select campaign.resource_name, campaign.name from campaign limit 1
+  EOQUERY
+
+  campaign_resource_name = res.first.campaign.resource_name
+
+  # note here, we have to pass the campaign (`camp`) as a block arg, because
```
mcloonan: Mention that the field mask is created and applied automatically based on what's set in the block.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -140,36 +131,73 @@ def service(name, version = default_api_version)
...
-            metadata: headers,
-            exception_transformer: ERROR_TRANSFORMER
-          )
+
+          if logger
+            logging_interceptor = Google::Ads::GoogleAds::LoggingInterceptor.new(logger)
+          end
+
+          if name.nil?
+            Factories::Services.new(
```
mcloonan: Should we instantiate one of these each time, or can we cache it and just reference the one we have?

Also, you need to build some infrastructure to support multiple versions of the factories at once, and use the correct version of them.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -140,36 +131,73 @@ def service(name, version = default_api_version)
...
 
         # Return a resource or common entity for the provided entity type. For
         # example, passing :Campaign will return an instantiated Campaign.
         #
         # Raises ArgumentError if no entity can be found for the provided type.
-        def resource(name, version = default_api_version)
-          lookup_util.resource(name, version)
+        def resource(name=nil, version=default_api_version)
+          if name.nil?
+            Google::Ads::GoogleAds::Factories::Resources
```
mcloonan: Same for all these; you need to have one set of factories per version and look up the correct one for the given version.

codegen/.gitignore
```diff
@@ -0,0 +1 @@
+examples.txt
```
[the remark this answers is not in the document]
samphippen: this is intentional to not commit, examples.txt is an RSpec state file that records info about tests as you're working on them. It's best practice to gitignore it.

codegen/src/filters.rb
```diff
@@ -0,0 +1,55 @@
...
+  }.partition { |klass, _|
+    !klass.name.end_with?("Operation")
+  }
+end
+
+Operation = Struct.new(
+  :operation,
+  :create_class,
+  :update_class,
+  :path
```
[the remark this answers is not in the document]
samphippen: not needed, removes just take a path.

codegen/src/filters.rb
```diff
@@ -0,0 +1,55 @@
...
+  potential_services.select { |service, _|
+    service.name.start_with?("Google::Ads::GoogleAds::V1")
+  }
+end
+
+def filter_resources_in_to_resources_and_operations(resources)
+  resources.reject { |klass, path|
+    klass.name.end_with?("Request", "Response")
+  }.partition { |klass, _|
+    !klass.name.end_with?("Operation")
```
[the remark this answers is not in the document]
samphippen: I like the return order as it stands now, so the ! is necessary, if you feel really strongly about this I'll change it, but otherwise I think this is good as is?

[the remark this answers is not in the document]
mcloonan: Yeah it's fine as is 👍

lib/google/ads/google_ads/google_ads_client.rb
samphippen: Creating one of these each time is probably safer for the time being, once this is more established and also once we investigate GRPC connection pooling further, we'll have a clearer idea of what to do here.

scripts/codegen.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euxo pipefail
```
mcloonan: Do we want to run this file automatically as part of the rake build?

PR #81, opened 2019-04-08
Validate all protos are loadable

samphippen:
Building on top of #80 this adds a script which loads every proto file and causes rake to terminate *after* patching all the proto files.

This way, we get validation that all the protos in the library are actually well formed and loadable.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -26,8 +26,15 @@ def apply_patches
   end
 end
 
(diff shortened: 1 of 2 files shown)


mcloonan:
Should this be part of a unit test rather than part of the release process? The unit tests get run as part of the release currently.

PR #92, opened 2019-04-12
Adding account linking example.

mcloonan:
Change-Id: I8634de007475a6b7ab29051855b21b076bc44e5a


examples/account_management/link_accounts.rb
```diff
@@ -0,0 +1,165 @@
...
+  client_link_operation['create'] = client_link
+
+  response = client_link_service.mutate_customer_client_link(
+    manager_customer_id,
+    client_link_operation,
+  )
+
+  client_link_resource_name = response.result.resource_name
+  puts "Extended an invitation from account #{manager_customer_id} to account " \
+      "#{client_customer_id} with resource name #{client_link_resource_name}."
```
jradcliff: nit: "client link resource name"
As written, could be misinterpreted as the customer resource name

examples/account_management/link_accounts.rb
```diff
@@ -0,0 +1,165 @@
...
+  manager_link_operation = client.operation(:CustomerManagerLink)
+  manager_link_operation['update'] = manager_link
+  manager_link_operation['update_mask'] = mask
+
+  response = manager_link_service.mutate_customer_manager_link(
+    client_customer_id,
+    [manager_link_operation],
+  )
+
+  puts "Accepted invitation with resource name #{response.results.first.resource_name}."
```
jradcliff: "Client accepted invitation..."

May be a bit repetitive, but I think this example can get confusing because of the different steps executing under different roles (manager or client).

examples/account_management/link_accounts.rb
```diff
@@ -0,0 +1,165 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This example demonstrates how to link two already existing accounts.
+
+require 'optparse'
+require 'google/ads/google_ads'
+
+def link_accounts(manager_customer_id, client_customer_id)
```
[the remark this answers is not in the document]
mcloonan: I went with "link_manager_to_client" to match the order of arguments.

PR #93, opened 2019-04-23
Add first pass of smart shopping ad

samphippen:


examples/shopping_ads/add_smart_shopping_ad.rb
```diff
@@ -0,0 +1,218 @@
...
+    campaign_resource_name,
+  )
+
+  add_smart_shopping_ad(
+    client,
+    customer_id,
+    ad_group_resource_name,
+  )
+
+  if ceate_default_listing_group
```
mcloonan: Typo. Can you make sure you test this code path?

examples/shopping_ads/add_smart_shopping_ad.rb
```diff
@@ -0,0 +1,218 @@
...
+      '--customer-id MERCHANT-CENTER-ACCOUNT-ID',
+      String,
+      'Merchant center account id'
+    ) do |v|
+      options[:merchant_center_account_id] = v
+    end
+
+
+    opts.on('-l', '--create-default-listing-group CREATE-DEFAULT-LISTING-GROUP', String, 'create default listing group') do |v|
+      options[:create_default_listing_group] = if v == "true"
```
mcloonan: WDYT of the convention established here for boolean command line flags: https://github.com/googleads/google-ads-ruby/blob/master/examples/shopping_ads/add_shopping_product_ad.rb#L215

samphippen: that'll create a truthy value if the user passes the literal string "false", the canonical correct way to do this in ruby these days is:

https://github.com/rails/rails/blob/master/activemodel/lib/active_model/type/boolean.rb

but I'm not sure we want to pull in activemodel or vendor this code in just for code samples. WDYT?

mcloonan: We definitely don't want to include any new dependencies for this. Can you update other examples that require true/false values to use whatever paradigm we go with here?

You can simplify this to: `options[:...] = v == "true"`

PR #94, opened 2019-04-23
Add an example to get campaigns by label

samphippen:


examples/campaign_management/get_campaigns_by_label.rb
```diff
@@ -0,0 +1,106 @@
...
+  EOD
+
+  ga_service = client.service(:GoogleAds)
+  response = ga_service.search(
+      customer_id,
+      query,
+      page_size: PAGE_SIZE)
+
+  response.each do |row|
+    puts sprintf("Campaign with ID %d and name '%s' was found.",
```
mcloonan: We've been using #{} rather than sprintf in newer examples. Eventually we'll go back through and update all of them to be consistent, but no reason to add more future work now.

examples/campaign_management/get_campaigns_by_label.rb
```diff
@@ -0,0 +1,106 @@
...
+
+require 'optparse'
+require 'google/ads/google_ads'
+
+def get_campaigns_by_label(customer_id, label_id)
+  # GoogleAdsClient will read a config file from
+  # ENV['HOME']/google_ads_config.rb when called without parameters
+  client = Google::Ads::GoogleAds::GoogleAdsClient.new
+
+  query = <<~EOD
```
mcloonan: Can we use <<~QUERY instead of EOD?

PR #98, opened 2019-04-24
Add an example to add campaign labels

samphippen:


examples/campaign_management/add_campaign_labels.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    opts.separator ''
+    opts.separator 'Options:'
+
+    opts.on('-C', '--customer-id CUSTOMER-ID', String, 'Customer ID') do |v|
+      options[:customer_id] = v
+    end
+
+    opts.on('-c', '--campaign-ids CAMPAIGN-IDS', String, 'Comma separated list of campaign ids') do |v|
+      options[:campaign_ids] = v.split(",").map(&:strip)
```
mcloonan: You should do the processing into an array after the command line processing in case someone specifies a comma separated list of campaigns in place of 'INSERT_CAMPAIGN_IDS_HERE' above rather than using command line arguments.

PR #100, opened 2019-04-29
add float

apneadiving:
Add missing float type cast


mcloonan:
I don't think any fields in the library protos actually use FloatValue. Did you run into an issue where you needed this?

apneadiving:
yes: Campaign Criterion / bid modifier
 https://developers.google.com/google-ads/api/reference/rpc/google.ads.googleads.v1.resources#campaigncriterion

mcloonan:
I think that field should be a DoubleValue, and I'm talking with the API eng team to confirm. However, for the time being, it is indeed a FloatValue, so I'll merge this PR. Thanks again!

PR #105, opened 2019-05-03
Allow users to pass custom authentication when creating a service.

### LINKED ISSUE #101: Allow user to pass through their own credentials
See for example: https://github.com/googleads/google-ads-ruby/blob/master/lib/google/ads/google_ads/v1/services/account_budget_service_client.rb#L78

The lib currently only accepts a refresh token and passes it along as part of the updater_proc, but we could allow the user to specify their credentials in any of the valid formats. We should add a config option to specify this with any custom value the user sets.

mcloonan:
Fixes #101

### DIFF
--- a/lib/google/ads/google_ads/google_ads_client.rb
+++ b/lib/google/ads/google_ads/google_ads_client.rb
@@ -141,7 +141,7 @@ module Google
             headers[:"login-customer-id"] = login_customer_id.to_s  # header values must be strings
           end
           return class_to_return.new(
-            credentials: get_updater_proc(),
+            credentials: get_credentials(),
             metadata: headers,
             exception_transformer: ERROR_TRANSFORMER
           )
@@ -229,6 +229,14 @@ module Google
           # If we don't find an error of the correct type, or if we run into an
           # error while processing, just throw the original.
           raise gax_error
+        end
+
+        def get_credentials()
+          if @config.authentication
+            @config.authentication
+          else
+            get_updater_proc()
+          end
         end
 
         # Provides the service a method by which to obtain an access token to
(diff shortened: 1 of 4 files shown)


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -141,7 +141,7 @@ def service(name, version = default_api_version)
             headers[:"login-customer-id"] = login_customer_id.to_s  # header values must be strings
           end
           return class_to_return.new(
-            credentials: get_updater_proc(),
+            credentials: get_credentials(),
```
samphippen: let's kill the parens here.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -231,6 +231,14 @@ def decode_partial_failure_error(pfe)
           raise gax_error
         end
 
+        def get_credentials()
+          if @config.authentication
+            @config.authentication
+          else
+            get_updater_proc()
```
samphippen: and here.

test/test_google_ads_client.rb
```diff
@@ -161,6 +162,29 @@ def test_logger_beats_log_target
       config.log_target = STDOUT
     end
 
-    assert_equal client.logger, logger
+    assert_equal(client.logger, logger)
+  end
+
+  def test_default_to_updater_proc
+    client = Google::Ads::GoogleAds::GoogleAdsClient.new() do |config|
```
samphippen: and here.

test/test_google_ads_client.rb
```diff
@@ -161,6 +162,29 @@ def test_logger_beats_log_target
...
+      config.client_secret = 'client_secret'
+      config.refresh_token = 'refresh_token'
+    end
+
+    credentials = client.get_credentials()
+    assert_instance_of(Proc, credentials)
+  end
+
+  def test_authentication_overrides_updater_proc
+    client = Google::Ads::GoogleAds::GoogleAdsClient.new() do |config|
```
samphippen: and here.

samphippen:
LGTM

PR #115, opened 2019-05-24
Factories deprecations code

samphippen:


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -148,6 +148,7 @@ def service(name=nil, version = default_api_version)
               exception_transformer: ERROR_TRANSFORMER
             )
           else
+            deprecate("Calling `google_ads_client.service(<ServiceType>)` is deprecated, please use `google_ads_client.service.<service_type>` instead")
```
mcloonan: We need this in resource, enum, and operation as well.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -300,6 +301,14 @@ def lookup_util
         def default_api_version
           Google::Ads::GoogleAds.default_api_version
         end
+
+        def deprecate(deprecation)
+          @deprecation ||= Deprecation.new(
```
mcloonan: Should we reinstantiate if config changed?

lib/google/ads/google_ads/deprecation.rb
```diff
@@ -0,0 +1,96 @@
...
+            Warning.warn("#{deprecation}. Called from: #{CallerFilter.first_non_google_ads_line}")
+          end
+        end
+
+        private
+
+        # Determines if we should issue a deprecation warning, silencing
+        # on each user call site after the first warning.
+        def should_warn?
+          true if @warn_on_all_deprecations
```
mcloonan: I don't think this returns since it's not the last statement in the method and it doesn't explicitly say return.

lib/google/ads/google_ads/deprecation.rb
```diff
@@ -0,0 +1,96 @@
...
+          increment = 5
+          loop do
+            stack = caller_locations(skip_frames, increment)
+            raise "No non-lib lines in stack" unless stack
+
+            line = stack.find { |l| l.path !~ IGNORE_REGEX }
+            return line.to_s if line
+
+            skip_frames += increment
+            increment   *= 2 # The choice of two here is arbitrary.
```
mcloonan: Some of this code is copied directly from CallerFilter in RSpec, so we need to attribute it correctly, assuming their license permits this type of use.

lib/google/ads/google_ads/google_ads_client.rb
samphippen: I think it's unlikely that a user will change their deprecation configs during mid execution, WDYT?

mcloonan: Also then it loses the state of where various warnings were created from. This should be fine.

lib/google/ads/google_ads/deprecation.rb
```diff
@@ -0,0 +1,100 @@
+module Google
+  module Ads
+    module GoogleAds
+      class Deprecation
+        Error = Class.new(StandardError)
```
mcloonan: Your commit message is "add tests" but I don't actually see any tests. Did you miss committing some?

lib/google/ads/google_ads/deprecation.rb
```diff
@@ -0,0 +1,100 @@
...
+        #                                           *
+        # 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25
+        #                 ^________________________^
+        #
+        # here, we've performed 15 lookup operations to find the item we're
+        # looking for, instead of 1000. The reason we use this chunk and doubling
+        # strategy is that pulling stack frames has some fixed overhead, and so
+        # in practice, doing a linear scan from frame 0 to the one we're interested
+        # in is also slower than this methodology, even though the doubling
+        # strategy may result in strictly more lookups thn necessary.
```
samphippen: @AnashOommen PTAL

lib/google/ads/google_ads/deprecation.rb
```diff
@@ -0,0 +1,100 @@
...
+            true
+          else
+            false
+          end
+        end
+      end
+
+      # @api private
+      # Portions of this implementation were copied from RSpec, under the
+      # MIT license: https://git.io/fjR7i
```
samphippen: @AnashOommen PTAL

AnashOommen: I *think* the use of this code is fine, but check with opensource-discuss@ before committing. I suspect that this code has to live in its own file, in a third_party folder instead of mixing it into your code.

Just curious, is there a ruby gem we can depend on instead for this snippet? That might be simpler.

samphippen: We're ok to use MIT licensed code according to opensource licensing.

samphippen:
I need to move the RSpec snippet to `third_party` however.

samphippen:
@AnashOommen @mcloonan I followed the third party procedure and rebased this to work as three commits that more logically follow the structure documented here: https://g3doc.corp.google.com/company/thirdparty/non-google3.md?cl=head please re-review.

PR #134, opened 2019-06-04
Extract service creation

samphippen:
This is a preparatory refactoring for allowing us to pass our own channels in to the code gax creates. We're doing this so that we can set the maximum message size on the channel to 64mb. To do this we need to pass a `GRPC::Core::Channel` in the position that `credentials` currently occupies. The problem that we have right now is that we've got two different ways of creating services. Until we've done some open source stuff, we can't merge that PR.

SO

this PR applies the "method object" refactoring to pull the (ever expanding) `Google::Ads::GoogleAdsClient#service` method out in to it's own object in the first commit, and then just applies the extract method refactoring over and over again. This basically is just to reduce the cognitive overhead of swapping in a grpc channel instead of an updater proc.

### DIFF
--- a/lib/google/ads/google_ads/google_ads_client.rb
+++ b/lib/google/ads/google_ads/google_ads_client.rb
@@ -52,7 +52,7 @@ require 'google/ads/google_ads/wrapper_util'
 require 'google/ads/google_ads/logging_interceptor'
 require 'google/ads/google_ads/factories'
 require 'google/ads/google_ads/errors'
-require 'google/ads/google_ads/lookup_service'
+require 'google/ads/google_ads/service_lookup'
 
 require 'google/gax'
 
@@ -116,7 +116,7 @@ module Google
         def service(name=nil, version = default_api_version)
           service_path = ENV['GOOGLEADS_SERVICE_PATH']
 
-          LookupService.new(
+          ServiceLookup.new(
             name,
             version,
             lookup_util,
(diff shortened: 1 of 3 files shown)


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -116,80 +116,15 @@ def configure(&block)
...
-            )
-          end
+          LookupService.new(
+            name,
+            version,
+            lookup_util,
+            service_path,
+            @logger,
+            @config,
+            get_updater_proc
```
mcloonan: This should be "get_credentials". It's actually a bug that it was using get_updater_proc and ignoring the @config.authentication. I came across this while working on service accounts.

samphippen: so this PR shouldn't change any behavior vs what's on top of master (that is we should not be conflating bug fixes and refactoring in PRs). I'd rather keep the current behavior (even if it's wrong, but passing the tests) and then do a second PR, on top of this one, to fix that behavior.

mcloonan: Ok, that works for me.

lib/google/ads/google_ads/lookup_service.rb
```diff
@@ -0,0 +1,154 @@
+require 'google/ads/google_ads/patch_lro_headers'
+
+module Google
+  module Ads
+    module GoogleAds
+      class LookupService
```
mcloonan: Can you name this "ServiceLookup" instead. "LookupService" reads to me like "a service for looking things up", rather than what this actually is, which is a util that looks up a service.

lib/google/ads/google_ads/lookup_service.rb
```diff
@@ -0,0 +1,154 @@
...
+        def initialize(name, version, lookup_util, service_path, logger, config, updater_proc)
+          @name = name
```
mcloonan: Can you put these in the same order as they are in the method signature.

PR #135, opened 2019-06-05
Custom channel

samphippen:


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -127,10 +128,33 @@ def service(name=nil, version = default_api_version)
...
+        def make_channel
+          # this fetch in to GoogleAdsServiceClient is pretty arbitrary, our
+          # protos all point at the same domain and port combo, so this just
+          # constructs it by pulling the values out of that client.
+          #
+          # V1 should also be pretty arbitrary here (that is we probably don't
+          # need to change this even with V2 protos present because the target
+          # doesn't change)
+          default_target = [
+            Google::Ads::GoogleAds::V1::Services::GoogleAdsServiceClient::SERVICE_ADDRESS,
```
mcloonan: I say just hard code the values again here rather than extracting them. They never should change.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -127,10 +128,33 @@ def service(name=nil, version = default_api_version)
             service_path,
             @logger,
             @config,
-            get_credentials,
+            make_channel,
```
mcloonan: I think you should update the ServiceLookup so this isn't called "updater_proc" there anymore.

PR #144, opened 2019-06-18
Adding inbound metadata size channel arg.

mcloonan:


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -138,7 +141,8 @@ def make_channel
           target = ENV.fetch('GOOGLEADS_SERVICE_PATH', default_target)
 
           channel_args = {
-            GRPC::Core::Channel::MAX_MESSAGE_LENGTH => 64*1000*1000
+            MAX_MESSAGE_LENGTH => 64*1000*1000,
+            MAX_METADATA_SIZE => 16*1000*1000,
```
AnashOommen: We use MB as 1024 * 1024, not 1000 * 1000.

PR #157, opened 2019-06-25
Replace all misc examples with factories

samphippen:


examples/misc/upload_image.rb
```diff
@@ -28,15 +28,15 @@ def upload_image(customer_id)
   # ENV['HOME']/google_ads_config.rb when called without parameters
   client = Google::Ads::GoogleAds::GoogleAdsClient.new
 
-  media_file = client.resource(:MediaFile)
+  media_file = client.resource.media_file
   media_file.type = :IMAGE
-  media_file.image = client.resource(:MediaImage)
+  media_file.image = client.resource.media_image
```
mcloonan: Can you actually use the block syntax to set the fields too when relevant?

PR #172, opened 2019-06-27
Adds very simple support for a HTTP proxy.

samphippen:
This works by setting the environment variable that [GRPC uses](https://github.com/grpc/grpc/blob/master/src/core/ext/filters/client_channel/http_proxy.cc)
when the user specifies it on their configuration file.

### DIFF
--- a/lib/google/ads/google_ads/config.rb
+++ b/lib/google/ads/google_ads/config.rb
@@ -16,6 +16,8 @@
 #
 # Configuration setup for and storage for the API.
 
+require 'uri'
+
 module Google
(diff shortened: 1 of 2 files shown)


lib/google/ads/google_ads/config.rb
```diff
@@ -62,6 +64,15 @@ def initialize(&block)
         def configure(&block)
           yield self
         end
+
+        def http_proxy=(uri)
```
mcloonan: Can you add a getter too in case people need to check which proxy they're using?

PR #178, opened 2019-07-02
Fix the error transformer for the v2 error

### LINKED ISSUE #177: GoogleAdsError never raise in v2
After the v2 update, the exception `Google::Ads::GoogleAds::Errors::GoogleAdsError` never raise.

samphippen:
Fixes #177

### DIFF
--- a/lib/google/ads/google_ads/service_lookup.rb
+++ b/lib/google/ads/google_ads/service_lookup.rb
@@ -80,17 +80,22 @@ module Google
           begin
             gax_error.status_details.each do |detail|
               # If there is an underlying GoogleAdsFailure, throw that one.
-              if detail.is_a?(
-                  Google::Ads::GoogleAds::V1::Errors::GoogleAdsFailure)
+              if [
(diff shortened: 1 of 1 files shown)


lib/google/ads/google_ads/service_lookup.rb
```diff
@@ -80,17 +80,22 @@ def validate_login_customer_id
           begin
             gax_error.status_details.each do |detail|
               # If there is an underlying GoogleAdsFailure, throw that one.
-              if detail.is_a?(
-                  Google::Ads::GoogleAds::V1::Errors::GoogleAdsFailure)
+              if [
+                Google::Ads::GoogleAds::V1::Errors::GoogleAdsFailure === detail,
```
mcloonan: We have a list of known versions. Can we soft-code this so we don't have to update it every time a new version is added or removed?

Also, even if we wanted to keep it this way, is there some advantage to using this wonky syntax over a normal `||` comparison?

PR #191, opened 2019-10-18
Enhance error inspection

devchas:


Rakefile
```diff
@@ -71,4 +71,4 @@ task :copy_third_party_code do |t|
 end
 
 task :build => [:codegen, :validate_protos]
-task :test => [:copy_third_party_code, :codegen]
+task :test
```
penelopezone: we don't want to actually commit this change I think.

test/test_errors.rb
```diff
@@ -56,6 +56,14 @@ def test_blank_code
     assert_equal({}, error_code)
   end
 
+  def test_inspect
+    error = build_error
+    error_code = Google::Ads::GoogleAds::Errors::GoogleAdsError.new(error)
+
+    expected = "#<#{error_code.class.name}: #{error_code.object_id} #{error.inspect}>"
+    assert_equal(expected, error_code.inspect)
```
penelopezone: 👏🏻

Rakefile
devchas: Reverted.

PR #195, opened 2019-12-04
Adding comment for ad group ad status

stevenberger:
- Adding comment about why ad group ad should be enabled for hotels, originally brought up in https://github.com/googleads/google-ads-dotnet/pull/123#discussion_r352711245


penelopezone:
Thanks!

PR #198, opened 2020-01-07
Implement changes for new versioning

penelopezone:


lib/google/ads/google_ads/version.rb
```diff
@@ -19,7 +19,7 @@
 module Google
   module Ads
     module GoogleAds
-      CLIENT_LIB_VERSION = '3.2.0'.freeze
+      CLIENT_LIB_VERSION = '4.0.0'.freeze
```
mcloonan: Also update the ChangeLog.

PR #214, opened 2020-07-08
Fixing performance of autoboxing fields

mcloonan:
The original implementation would scan the list of fields and repeated fields on each instantiation, looking for ones that match what the user passed in. The new implementation stores the data for fields and repeated fields in a hash for quick lookup on kwarg name, and then only looks at the ones relevant to the user's request.

Also if the user passes no kwargs, it skips the whole thing.

### DIFF
--- a/lib/google/ads/google_ads/autoboxing_fields.rb
+++ b/lib/google/ads/google_ads/autoboxing_fields.rb
@@ -37,17 +37,19 @@ module Google
         def self.patch_constructor_for_autoboxing(fields, repeated_fields, klass_to_patch)
           orig_initialize = klass_to_patch.instance_method(:initialize)
           klass_to_patch.instance_eval do
+            fields_map = fields.map { |field| [field.name.to_sym, field] }.to_h
+            repeated_fields_map = repeated_fields.map { |field| [field.name.to_sym, field] }.to_h
             define_method(:initialize) do |**kwargs|
               new_kwargs = kwargs.dup
-              fields.select { |x| new_kwargs.include?(x.name.to_sym) }.each do |field|
-                value = new_kwargs.fetch(field.name.to_sym)
-                new_kwargs[field.name.to_sym] = AutoboxingMappings.wrapped_mapping(field.subtype.msgclass).call(value)
-              end
-
-              repeated_fields.select { |x| new_kwargs.include?(x.name.to_sym) }.each do |field|
-                value = new_kwargs.fetch(field.name.to_sym)
-                mapping = AutoboxingMappings.wrapped_mapping(field.subtype.msgclass)
-                new_kwargs[field.name.to_sym] = value.map { |x| mapping.call(x) }
+              if !kwargs.empty?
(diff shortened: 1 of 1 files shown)


lib/google/ads/google_ads/autoboxing_fields.rb
```diff
@@ -37,17 +37,19 @@ def self.is_value_field?(class_name)
...
-                value = new_kwargs.fetch(field.name.to_sym)
-                new_kwargs[field.name.to_sym] = AutoboxingMappings.wrapped_mapping(field.subtype.msgclass).call(value)
-              end
-
-              repeated_fields.select { |x| new_kwargs.include?(x.name.to_sym) }.each do |field|
-                value = new_kwargs.fetch(field.name.to_sym)
-                mapping = AutoboxingMappings.wrapped_mapping(field.subtype.msgclass)
-                new_kwargs[field.name.to_sym] = value.map { |x| mapping.call(x) }
+              if !kwargs.empty?
+                kwargs.each do |kwarg, value|
```
jradcliff, line 45: I noticed you aren't calling `kwarg.to_sym` here, but the old code did. Is that an issue?

lib/google/ads/google_ads/autoboxing_fields.rb
```diff
@@ -37,17 +37,19 @@ def self.is_value_field?(class_name)
         def self.patch_constructor_for_autoboxing(fields, repeated_fields, klass_to_patch)
```
AnashOommen, line 37: Does this method need a thread locking, to make sure field_map and repeated_fields_map don't end up in an indeterministic state in a multithreaded env?

mcloonan: If they do, then that's outside the scope of this PR. I can look into that separately.

lib/google/ads/google_ads/autoboxing_fields.rb
mcloonan: I don't think we ever called to_sym on the kwarg provided from the user. Those should all be symbols because of how the language works.

We did call to_sym on the field name from the field descriptor, which is still around but done once at the beginning when I generate the map.

lib/google/ads/google_ads/autoboxing_fields.rb
AnashOommen: I think this is _mostly_ fine, since Ruby doesn't have parallelism, just concurrency. My knowledge on Ruby is dated.

Here's some reading material:

https://medium.com/@zanker/ruby-vm-and-how-apps-break-part-1-81a309384f39
https://medium.com/better-programming/scalable-ruby-concurrency-and-parallelism-explained-68b09a7aeb53

lib/google/ads/google_ads/autoboxing_fields.rb
jradcliff: Sorry, I somehow missed the `to_sym` call above. LGTM.

PR #221, opened 2020-09-07
Add 4 code examples

xflesym:
- approve_merchant_center_link
- add_customer_match_user_list
- upload_call_conversion
- search_for_language_and_carrier_constants

### DIFF
--- a/examples/remarketing/add_customer_match_user_list.rb
+++ b/examples/remarketing/add_customer_match_user_list.rb
@@ -1,0 +1,244 @@
+#!/usr/bin/env ruby
+# Encoding: utf-8
+#
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This example uses Customer Match to create a new user list (a.k.a. audience)
+# and adds users to it.
+#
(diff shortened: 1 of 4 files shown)


examples/remarketing/add_customer_match_user_list.rb
```diff
@@ -0,0 +1,244 @@
...
+  )
+
+  # Prints out some information about the newly created user list.
+  resource_name = response.results.first.resource_name
+  puts "User list with resource name #{resource_name} was created."
+
+  resource_name
+end
+
+def add_users_to_customer_match_user_list(client, customer_id, user_list)
```
mcloonan, line 68: I'm not sure how we're adding users without passing in some information about the users. I think this should be refactored to take a list of emails and we can default populate it with a few fake @example.com ones.

examples/remarketing/upload_call_conversion.rb
```diff
@@ -0,0 +1,160 @@
...
+  options = {}
+  # The following parameter(s) should be provided to run the example. You can
+  # either specify these by changing the INSERT_XXX_ID_HERE values below, or on
+  # the command line.
+  #
+  # Parameters passed on the command line will override any parameters set in
+  # code.
+  #
+  # Running the example with -h will print the command line usage.
+  options[:customer_id] = 'INSERT_CUSTOMER_ID_HERE'
```
mcloonan, line 95: Missing other default values.

mcloonan (changes requested):

AnashOommen (approved):

examples/remarketing/add_customer_match_user_list.rb
xflesym: The example adds two users, one using email address, the other one using physical address (first name, last name, country code, postal code). It might be too many default values to add, what do you think?

mcloonan: SGTM to leave this as is.

examples/remarketing/upload_call_conversion.rb
xflesym: per above, only CUSTOMER_ID is needed here.

PR #224, opened 2020-09-08
Add add_complete_campaigns_using_batch_job.rb code example

xflesym:


examples/campaign_management/add_complete_campaigns_using_batch_job.rb
```diff
@@ -0,0 +1,370 @@
...
+
+require 'optparse'
+require 'google/ads/google_ads'
+
+def add_complete_campaigns_using_batch_job(customer_id)
+  # GoogleAdsClient will read a config file from
+  # ENV['HOME']/google_ads_config.rb when called without parameters
+  client = Google::Ads::GoogleAds::GoogleAdsClient.new
+
+  batch_job_service_client = client.service.batch_job
```
mcloonan, line 29: Just call this "batch_job_service"

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
```diff
@@ -0,0 +1,370 @@
...
+  poll_batch_job(operation_response)
+
+  fetch_and_print_results(batch_job_service_client, batch_job_resource_name)
+end
+
+# Creates a new batch job for the specified customer ID.
+def create_batch_job(client, batch_job_service_client, customer_id)
+  # Creates a batch job operation to create a new batch job.
+  operation = client.operation.create_resource.batch_job do |bj|
+  end
```
mcloonan, line 58: I don't think you need to use a block at all if you're not doing anything with it. Just client.operation.create_resource.batch_job should be sufficient.

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
```diff
@@ -0,0 +1,370 @@
...
+
+# Builds all operations for creating a complete campaign and return an array of
+# their corresponding mutate operations.
+def build_all_operations(client, customer_id)
+  mutate_operations = []
+
+  # Creates a new campaign budget operation and add it to the array of mutate
+  # operations.
+  campaign_budget_operation = build_campaign_budget_operation(client, customer_id)
+  mop = client.operation.mutate
```
mcloonan, line 135: "mop" is a word that has nothing to do with what you're doing. I think you were going for something like "mutate_op", but maybe spell it out so it doesn't sound silly to native speakers :)

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
```diff
@@ -0,0 +1,370 @@
...
+    mutate_operations << mop
+  end
+
+  # Creates new campaign criterion operations and adds them to the array of
+  # mutate operations.
+  campaign_criterion_operations = build_campaign_criterion_operations(
+    client, campaign_operations)
+  campaign_criterion_operations.each do |op|
+    mop = client.operation.mutate
+    mop.campaign_criterion_operation = op
```
mcloonan, line 155: Does "client.operation.mutate" not support a block like "client.resource.whatever" would? If it does support that, please use that syntax. If it doesn't, please file an issue for our library to add that functionality and then update this example to use that.

mcloonan (changes requested):

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
```diff
@@ -0,0 +1,370 @@
...
+      end
+    end
+  end
+
+  operations
+end
+
+# Builds new ad group criterion operations for creating keywords. 50% of
+# keywords are created with some invalid characters to demonstrate how
+# MutateJobService returns information about such errors.
```
PierrickVoulet, line 271: Please replace `MutateJobService` with [`BatchJobService`](https://developers.google.com/google-ads/api/reference/rpc/v5/BatchJobService)

PierrickVoulet (changes requested):

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
xflesym: The operation would be empty if I don't use a block here. Do you want me to file a bug?
(I've now added a comment here to explain the empty block)

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
xflesym: done

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
xflesym: Makes sense :) Changed to mutate_op

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
xflesym: Good catch. Changed to BatchJobService.

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
xflesym: It does not support block syntax. I'll file an issue to add that and update example later.

PierrickVoulet (approved):

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
mcloonan: Yeah, you shouldn't need to pass an empty block, so that seems like a bug to me.

mcloonan (approved):

examples/campaign_management/add_complete_campaigns_using_batch_job.rb
xflesym: Created https://github.com/googleads/google-ads-ruby/issues/227 to track it

PR #232, opened 2020-09-15
add parallel_report_download.rb code example

xflesym:


examples/reporting/parallel_report_download.rb
```diff
@@ -0,0 +1,186 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Shows how to download a set of reports from a list of accounts in parallel.
+# If you need to obtain a list of accounts, please see get_account_hierarchy.rb
+# or list_accessible_customers.rb examples.";
```
mcloonan, line 20: Extra punctuation at the end.

examples/reporting/parallel_report_download.rb
```diff
@@ -0,0 +1,186 @@
...
+        end
+      end
+    end
+  end
+
+  puts "Awaiting results from #{threads.size} report download requests..."
+
+  # Waits for all pending requests to the current set of customer IDs
+  # to complete.
+  threads.each { |aThread|  aThread.join }
```
mcloonan, line 107: Just call this "thread" instead of "aThread"

examples/reporting/parallel_report_download.rb
```diff
@@ -0,0 +1,186 @@
...
+  # via a manager account.
+  client.configure do |config|
+    if login_customer_id
+      config.login_customer_id = login_customer_id.tr("-", "").to_i
+    end
+  end
+
+  ga_service = client.service.google_ads
+
+  query_list = {}
```
mcloonan, line 41: Based on the way you're using query_list below, this should actually be an array that stores 2 pieces of information per element, rather than a map of key->value. Can you refactor it to be an array instead?

examples/reporting/parallel_report_download.rb
```diff
@@ -0,0 +1,186 @@
...
+
+  puts "Awaiting results from #{threads.size} report download requests..."
+
+  # Waits for all pending requests to the current set of customer IDs
+  # to complete.
+  threads.each { |aThread|  aThread.join }
+
+  puts 'Download completed, results:'
+  puts 'Successful reports:'
+  while !reports_succeeded.empty? do
```
mcloonan, line 111: Can you juste do reports_succeeded.each do |result| instead? And similar below?

In Ruby prefer to use these kinds of iterators rather than loops like you'd find in most other programming languages when possible.

examples/reporting/parallel_report_download.rb
```diff
@@ -0,0 +1,186 @@
...
+    'INSERT_CUSTOMER_ID_2_HERE',
+  ]
+
+  OptionParser.new do |opts|
+    opts.banner = sprintf('Usage: ruby %s [options]', File.basename(__FILE__))
+
+    opts.separator ''
+    opts.separator 'Options:'
+
+    opts.on('-C', '--customer-ids CUSTOMER-IDS', String, 'Customer ID') do |v|
```
mcloonan, line 149: Add a little more information to the help text explaining how/why multiple customer IDs are implemented.

mcloonan (changes requested):

examples/reporting/parallel_report_download.rb
xflesym: Fixed.

examples/reporting/parallel_report_download.rb
xflesym: Fixed.

examples/reporting/parallel_report_download.rb
xflesym: Added some more help text to explain, for customer ID and login customer ID as well.

examples/reporting/parallel_report_download.rb
xflesym: Changed to an array.

examples/reporting/parallel_report_download.rb
xflesym: Unfortunately the Queue class in Ruby doesn't have a `each` method. So we have to use loops here.

mcloonan: Ok, that makes sense. Can you add a comment when you create these that using a Queue is to make sure it's thread-safe, and that an array wouldn't be.

mcloonan (approved):

examples/reporting/parallel_report_download.rb
xflesym: Yup sure. I've added comments.

PR #234, opened 2020-09-16
Added get_ad_group_criterion_cpc_bid_simulations.rb

wayneluzw:
This example gets all available ad group criterion CPC bid simulations for a given ad group.

### DIFF
--- a/examples/planning/get_ad_group_criterion_cpc_bid_simulations.rb
+++ b/examples/planning/get_ad_group_criterion_cpc_bid_simulations.rb
@@ -39,7 +39,7 @@ def get_ad_group_criterion_cpc_bid_simulations(customer_id, ad_group_id)
 
   responses = client.service.google_ads.search_stream(
     customer_id: customer_id,
-    query: query
+    query: query,
   )
 
   responses.each do |response|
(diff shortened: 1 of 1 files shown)


examples/planning/get_ad_group_criterion_cpc_bid_simulations.rb
```diff
@@ -0,0 +1,125 @@
...
+           ad_group_criterion_simulation.end_date,
+           ad_group_criterion_simulation.cpc_bid_point_list.points
+    FROM ad_group_criterion_simulation
+    WHERE ad_group_criterion_simulation.type = CPC_BID 
+    AND ad_group_criterion_simulation.ad_group_id = #{ad_group_id}
+  QUERY
+
+  responses = client.service.google_ads.search_stream(
+    customer_id: customer_id,
+    query: query
```
xflesym, line 42: nit: we usually put a comma at the end of last parameter too, so that future changes will have the minimal lines of diff

xflesym (approved):
lgtm with a minor comment

examples/planning/get_ad_group_criterion_cpc_bid_simulations.rb
wayneluzw: Sounds good, thanks.

PR #238, opened 2020-09-21
Add two code examples

xflesym:
add_ad_customizer.rb
upload_store_sales_transactions.rb

### DIFF
--- a/examples/advanced_operations/add_ad_customizer.rb
+++ b/examples/advanced_operations/add_ad_customizer.rb
@@ -1,0 +1,383 @@
+#!/usr/bin/env ruby
+# Encoding: utf-8
+#
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Adds an ad customizer feed and associates it with the customer. Then it adds
+# an ad that uses the feed to populate dynamic data.
+
+require 'date'
+require 'google/ads/google_ads'
+require 'optparse'
+
+def add_ad_customizer(customer_id, ad_group_ids)
+  # GoogleAdsClient will read a config file from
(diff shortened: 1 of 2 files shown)


AnashOommen (approved):

examples/advanced_operations/add_ad_customizer.rb
```diff
@@ -0,0 +1,383 @@
...
+  operation = client.operation.create_resource.feed_mapping do |fm|
+    fm.placeholder_type = :AD_CUSTOMIZER
+    fm.feed = ad_customizer_feed_resource_name
+    # Maps the feed attribute IDs to the field ID constants.
+    fm.attribute_field_mappings << client.resource.attribute_field_mapping do |af|
+      af.feed_attribute_id = ad_customizer_feed_attributes["Name"]
+      af.ad_customizer_field = :STRING
+    end
+    fm.attribute_field_mappings << client.resource.attribute_field_mapping do |af|
+      af.feed_attribute_id = ad_customizer_feed_attributes["Price"]
```
mcloonan, line 156: You should use symbols instead of strings as hash keys whenever you can. Strings in Ruby are mutable, so every time you hard code a string it uses new memory. Symbols are immutable, so multiple of the same symbol use the same memory. You can use "to_sym" to convert a string to a symbol.

examples/advanced_operations/add_ad_customizer.rb
```diff
@@ -0,0 +1,383 @@
...
+# customizing a specific ad group.
+def create_feed_item_targets(
+  client,
+  customer_id,
+  ad_group_ids,
+  feed_item_resource_names)
+  # Bind each feed item to a specific ad group to make sure it will only be
+  # used to customize ads inside that ad group; using the feed item elsewhere
+  # will result in an error.
+  for i in 0..(feed_item_resource_names.size - 1)
```
mcloonan, line 256: (feed_item_resource_names.size - 1).times do |i|

mcloonan: or (0..(feed_item_resource_names.size - 1)).each do |i|

mcloonan: The point is just prefer to use an iterator rather than a normal "loop" like in other languages.

mcloonan (changes requested):

examples/advanced_operations/add_ad_customizer.rb
xflesym: Changed to using `times`.

examples/advanced_operations/add_ad_customizer.rb
xflesym: Switched to using symbols.

mcloonan (approved):

PR #240, opened 2020-09-22
add 2 code examples: reject_merchant_center_link.rb and remove_feed_items.rb

xflesym:
add 2 code examples: reject_merchant_center_link.rb and remove_feed_items.rb


examples/account_management/reject_merchant_center_link.rb
```diff
@@ -0,0 +1,144 @@
...
+  # Rejects a pending link request or unlinks an enabled link for a Google Ads
+  # account with customer_id from a Merchant Center account with
+  # merchant_center_account_id.
+  response = merchant_center_link_service.list_merchant_center_links(
+    customer_id: customer_id,
+  )
+  puts "#{response.merchant_center_links.size} Merchant Center link(s) found " \
+    "with the following details:"
+
+  response.merchant_center_links.each do |link|
```
mcloonan, line 45: You could probably use response.merchant_center_links.detect {|link| link.id == merchange_center_account_id.to_i} to extract the one value you're actually interested in, and then act on that one, rather than needing to use a break below.

mcloonan (changes requested):

examples/account_management/reject_merchant_center_link.rb
xflesym: Changed to using detect. Thanks for the suggestion!

mcloonan:
Looks like there's a merge issue going on since this was created before the v6 release. You can try to resolve or just create a new PR with the new files if that's easier.

xflesym:
Create a new PR https://github.com/googleads/google-ads-ruby/pull/264
Closing this one.

PR #249, opened 2020-11-19
support loading configuration from environment variables

xflesym:


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
           yield @config
         end
 
+        def configure_using_environment_variables
```
mcloonan, line 100: Can we call this "load_environment_config" instead? I think "configure" implies that the user should be passing something, but "load" implies that there is some information already stored somewhere, which is more accurate.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
...
+          # Client library-generic variables
```
mcloonan, line 101: Just "Generic variables",

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
...
+          @config.keyfile = ENV.fetch("GOOGLE_ADS_JSON_KEY_FILE_PATH", @config.keyfile)
+          @config.impersonate = ENV.fetch("GOOGLE_ADS_IMPERSONATED_EMAIL", @config.impersonate)
+          @config.developer_token = ENV.fetch("GOOGLE_ADS_DEVELOPER_TOKEN", @config.developer_token)
+          @config.login_customer_id = ENV.fetch("GOOGLE_ADS_LOGIN_CUSTOMER_ID", @config.login_customer_id)
+          @config.linked_customer_id = ENV.fetch("GOOGLE_ADS_LINKED_CUSTOMER_ID", @config.linked_customer_id)
+
+          # Client library-specific variables
+          @config.treat_deprecation_warnings_as_errors = ENV.fetch("GOOGLE_ADS_RUBY_TREAT_DEPRECATION_WARNINGS_AS_ERRORS", @config.treat_deprecation_warnings_as_errors)
+          @config.warn_on_all_deprecations = ENV.fetch("GOOGLE_ADS_RUBY_WARN_ON_ALL_DEPRECATIONS", @config.warn_on_all_deprecations)
+          @config.authentication = ENV.fetch("GOOGLE_ADS_RUBY_AUTHENTICATION", @config.authentication)
```
mcloonan, line 113: This one actually can't really be configured via environment variable, since it requires one of various types of Ruby object. I don't think we should include this one. Anyone who wants to specify this one must do it in code.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
...
+          @config.log_level = ENV.fetch("GOOGLE_ADS_RUBY_LOG_LEVEL", @config.log_level)
+          @config.log_target = ENV.fetch("GOOGLE_ADS_RUBY_LOG_TARGET", @config.log_target)
+          @config.logger = ENV.fetch("GOOGLE_ADS_RUBY_LOGGER", @config.logger)
```
mcloonan, line 116: Same here. You can't specify this one as a string; it must be a Ruby object. So don't include it here.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
...
+          @config.impersonate = ENV.fetch("GOOGLE_ADS_IMPERSONATED_EMAIL", @config.impersonate)
+          @config.developer_token = ENV.fetch("GOOGLE_ADS_DEVELOPER_TOKEN", @config.developer_token)
+          @config.login_customer_id = ENV.fetch("GOOGLE_ADS_LOGIN_CUSTOMER_ID", @config.login_customer_id)
+          @config.linked_customer_id = ENV.fetch("GOOGLE_ADS_LINKED_CUSTOMER_ID", @config.linked_customer_id)
+
+          # Client library-specific variables
+          @config.treat_deprecation_warnings_as_errors = ENV.fetch("GOOGLE_ADS_RUBY_TREAT_DEPRECATION_WARNINGS_AS_ERRORS", @config.treat_deprecation_warnings_as_errors)
+          @config.warn_on_all_deprecations = ENV.fetch("GOOGLE_ADS_RUBY_WARN_ON_ALL_DEPRECATIONS", @config.warn_on_all_deprecations)
+          @config.authentication = ENV.fetch("GOOGLE_ADS_RUBY_AUTHENTICATION", @config.authentication)
+          @config.log_level = ENV.fetch("GOOGLE_ADS_RUBY_LOG_LEVEL", @config.log_level)
```
mcloonan, line 114: You may have to convert this one to a symbol for it to actually work properly.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
...
+          @config.log_target = ENV.fetch("GOOGLE_ADS_RUBY_LOG_TARGET", @config.log_target)
```
mcloonan, line 115: I'm pretty sure this one won't work either, since the main thing you may specify is a Ruby constant like STDOUT or an open file. Can you write some test cases for the Ruby-library specific ones to ensure they each work as intended?

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -96,6 +97,25 @@ def configure(&block)
...
+          # Client library-generic variables
+          @config.client_id = ENV.fetch("GOOGLE_ADS_CLIENT_ID", @config.client_id)
+          @config.client_secret = ENV.fetch("GOOGLE_ADS_CLIENT_SECRET", @config.client_secret)
+          @config.keyfile = ENV.fetch("GOOGLE_ADS_JSON_KEY_FILE_PATH", @config.keyfile)
+          @config.impersonate = ENV.fetch("GOOGLE_ADS_IMPERSONATED_EMAIL", @config.impersonate)
+          @config.developer_token = ENV.fetch("GOOGLE_ADS_DEVELOPER_TOKEN", @config.developer_token)
+          @config.login_customer_id = ENV.fetch("GOOGLE_ADS_LOGIN_CUSTOMER_ID", @config.login_customer_id)
+          @config.linked_customer_id = ENV.fetch("GOOGLE_ADS_LINKED_CUSTOMER_ID", @config.linked_customer_id)
+
+          # Client library-specific variables
```
mcloonan, line 110: Can you include one for the http_proxy?

test/test_google_ads_client.rb
```diff
@@ -228,4 +228,21 @@ def test_authentication_overrides_updater_proc_and_keyfile
...
+
+  def test_load_environment_variables
+    # Test config file path
+    ENV.store("GOOGLE_ADS_CONFIGURATION_FILE_PATH", "test/fixtures/config.rb")
+    client = Google::Ads::GoogleAds::GoogleAdsClient.new
+    assert_equal(client.config.developer_token, "INSERT_DEVELOPER_TOKEN_HERE")
+    assert_equal(client.config.client_id, "INSERT_CLIENT_ID_HERE")
+    assert_equal(client.config.client_secret, "INSERT_CLIENT_SECRET_HERE")
+    # Test environment variable overrides
+    ENV.store("GOOGLE_ADS_DEVELOPER_TOKEN", "XYZ")
```
mcloonan, line 240: Move these to the top so that we can test they're not loaded until the user manually calls the method as well.

mcloonan (changes requested):
Awesome, this looks great! Just a couple minor changes.

lib/google/ads/google_ads/google_ads_client.rb
PierrickVoulet: Another option would be something like "load_from_environment_variables"

lib/google/ads/google_ads/google_ads_client.rb
PierrickVoulet, line 101: `GOOGLE_ADS_REFRESH_TOKEN` is missing

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -117,7 +137,7 @@ def endpoint
 
         def target
           default_target = "googleads.googleapis.com:443"
-          target = ENV.fetch('GOOGLEADS_SERVICE_PATH', default_target)
```
PierrickVoulet, line 120: You may want to keep the old one for a while and flag it as deprecated.

lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -117,7 +137,7 @@ def endpoint
...
+          target = ENV.fetch('GOOGLE_ADS_ENDPOINT', default_target)
```
PierrickVoulet, line 140: This seems to always be loaded even when `configure_using_environment_variables` is never called. You may need to redesign this to use a new configuration setting instead.

test/fixtures/config.rb
```diff
@@ -0,0 +1,84 @@
...
+  # refresh_token fields above.
+  # c.authentication = INSERT_AUTHENTICATION_METHOD_HERE
+
+  # Required for manager accounts only: Specify the login customer ID used to
+  # authenticate API calls. This will be the customer ID of the authenticated
+  # manager account. If you need to use different values for this field, then
+  # make sure fetch a new copy of the service after each time you change the
+  # value.
+  # c.login_customer_id = 'INSERT_LOGIN_CUSTOMER_ID_HERE'
+
```
PierrickVoulet, line 62: `linked_customer_id` is missing in this config file example

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Since this is an internal setting, external developers wouldn't use the old one any ways, I prefer replacing it rather than keeping it, unless there is strong reason not to do so. WDYT?

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Added.

test/test_google_ads_client.rb
xflesym: Done.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Done.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Added test cases for Ruby-specific ones.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Currently we use string like 'WARN' in our config file, so string should be fine here.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Removed the ones that don't work with strings.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Removed.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Done.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Makes sense. Changed it to load_from_environment_variables

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Changed to a new design to read from config setting.

test/fixtures/config.rb
xflesym: Good catch, thanks! Added it in example config.

lib/google/ads/google_ads/google_ads_client.rb
PierrickVoulet: SGTM.

PierrickVoulet (approved):

lib/google/ads/google_ads/google_ads_client.rb
mcloonan: +1 on just changing it right away; Fei and I are the only users of this setting.

lib/google/ads/google_ads/google_ads_client.rb
mcloonan: This is a method on the GoogleAdsClient, though, so "client.load_from_environment_variables" actually doesn't make as much sense as "client.load_config_from_environment_variables", but that's super long, which is why I originally proposed "client.load_environment_config" which is more reasonable to type. We have to specify *what* we're loading because the client itself isn't loading, one of its sub-pieces is.

lib/google/ads/google_ads/google_ads_client.rb
mcloonan: This one should *always* load if specified, regardless of config setting. It's only for internal development, so it won't cause any issues for end users.

mcloonan (changes requested):
Added a few more comments.

lib/google/ads/google_ads/google_ads_client.rb
PierrickVoulet: Good point, @mcloonan . A few things:

1. It makes better sense to use `GOOGLE_ADS_CONFIGURATION_FILE_PATH` to switch between configurations instead of loading the endpoint all the time.
1. @wfansh and I were discussing the possibility to add the "load from environment variables" call to all code examples for your use case but also our functional tests. We could make this call optional with an additional argument in all code examples. @AnashOommen WDYT?

mcloonan: 1. We specifically chose not to allow setting the endpoint via the config so that users don't think it's something they should do. This is something that only maintainers of the library can ever use, so it doesn't need to be "user-friendly" in the normal sense.
2. I'm strongly opposed to modifying every example for any reason at this point.

mcloonan: After some discussion, this looks good as is. Allow configuring the endpoint in the config and via environment variable only when explicitly loading the env config.

I'm still opposed to modifying every example, though :)

lib/google/ads/google_ads/config.rb
```diff
@@ -31,6 +31,7 @@ class Config
 
         attr_accessor :developer_token
         attr_accessor :login_customer_id
+        attr_accessor :linked_customer_id
```
mcloonan, line 34: While you're implementing this, could you also actually update the code to support it? It should be treated pretty much exactly the same as login_customer_id, passed along to the API as a header.

xflesym: Yup sure. Added the support.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: Seems like we have a consensus :) I'll resolve this thread and keep the current implementation.

lib/google/ads/google_ads/google_ads_client.rb
xflesym: load_environment_config SGTM. Changed it accordingly. Thanks.

xflesym:
Small change to method name.
Added header support for linked_customer_id.

mcloonan (approved):

PR #258, opened 2020-11-25
Add note about technical comments in code examples

PierrickVoulet:


README.md
```diff
@@ -113,6 +113,9 @@ To fetch a specific service, for example CampaignService:
 See the provided [examples][] for more detailed demonstrations of how to use the
 library.
 
+**Note:** From the source code, you can ignore comments with `[START...]` and
+`[END...]` in them-they are specifically used by Google.
```
mcloonan, line 117: Please use an em dash (—) instead of a hyphen (-).

README.md
```diff
@@ -113,6 +113,9 @@ To fetch a specific service, for example CampaignService:
 See the provided [examples][] for more detailed demonstrations of how to use the
 library.
 
+**Note:** From the source code, you can ignore comments with `[START...]` and
```
mcloonan, line 116: Can we reword as "You can ignore comments with .... in examples—they are specifically used by Google." I think that flows better.

Also, can we be any more specific about what we're using it for? It's not malicious, so being more open is probably better.

mcloonan (changes requested):

README.md
PierrickVoulet: I reworded as suggested, thanks!

When it comes to adding extra details about what these are used for though, we discussed that with Q & Nick and we decided to not do it. Let's discuss this offline if you strongly believe that it is required.

mcloonan (approved):

PR #260, opened 2020-12-01
fix block syntax for operations (issue #225 and #227)

### LINKED ISSUE #225: Block syntax support for client.operation.mutate
Currently client.operation.mutate does not support block syntax, i.e.
mutate = client.operation.mutate do |m|
  # do something with m
end

We need to add support for this syntax, and update code examples to use this.
Example(s) to be updated:
add_complete_campaigns_using_batch_job.rb

### LINKED ISSUE #227: client.operation.create_resource.batch_job does not create operation correctly
Currently we need an empty do block for the batch_job operation to be created correctly, e.g.:

operation = client.operation.create_resource.batch_job do |job|
end

We should not need that. Just operation = **client.operation.create_resource.batch_job** should do it.
Please investigate and fix.

xflesym:
- support block syntax for client.operation.operation_name
- support direct creation without block for client.operation.create_resource.resource_type

resolves #225, resolves #227

### DIFF
--- a/codegen/templates/operations.rb.erb
+++ b/codegen/templates/operations.rb.erb
@@ -10,7 +10,13 @@ module Google
             # @return [<%= operation.operation.name %>] the operation
             def self.<%= operation.operation.name.split("::").last.underscore.gsub("_operation", "") %>
               require "<%= operation.path %>"
-              <%= operation.operation.name %>.new
+              if block_given?
+                op = <%= operation.operation.name %>.new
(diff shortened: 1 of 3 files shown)


mcloonan (commented):
Can you add a test for this also?

xflesym:
Added a test case for operation/resource creation.

mcloonan (approved):

PR #276, opened 2020-12-09
add code example get_invoices.rb

xflesym:


examples/billing/get_invoices.rb
```diff
@@ -0,0 +1,148 @@
...
+  # Gets the date one month before now.
+  last_month = Date.today.prev_month
+
+  # Issues a request to list invoices.
+  response = client.service.invoice.list_invoices(
+    customer_id: customer_id,
+    billing_setup: client.path.billing_setup(customer_id, billing_setup_id),
+    # The year needs to be 2019 or later.
+    issue_year: last_month.year.to_s,
+    issue_month: last_month.strftime("%B").upcase.to_sym,
```
mcloonan, line 39: Add a comment explaining what %B does.

examples/billing/get_invoices.rb
```diff
@@ -0,0 +1,148 @@
...
+  client = Google::Ads::GoogleAds::GoogleAdsClient.new
+
+  # Gets the date one month before now.
+  last_month = Date.today.prev_month
+
+  # Issues a request to list invoices.
+  response = client.service.invoice.list_invoices(
+    customer_id: customer_id,
+    billing_setup: client.path.billing_setup(customer_id, billing_setup_id),
+    # The year needs to be 2019 or later.
```
mcloonan, line 37: Is this hard-coded as 2019 or is it a rolling window some distance in the past?

examples/billing/get_invoices.rb
```diff
@@ -0,0 +1,148 @@
...
+    issue_year: last_month.year.to_s,
+    issue_month: last_month.strftime("%B").upcase.to_sym,
+  )
+
+  # Iterates over all invoices retrieved and prints their information.
+  response.invoices.each do |invoice|
+    puts "- Found the invoice '#{invoice.resource_name}'" \
+      " ID (also known as Invoice Number): '#{invoice.id}'\n" \
```
mcloonan, line 45: This giant block of text would probably be easier to read for a human and easier to edit for future maintainers if we use the heredoc format like we do for queries in other examples.

```
puts <<~OUTPUT
  ...
OUTPUT 
```

Then you could remove all the newlines, quotes, and backslashes.

mcloonan (changes requested):

examples/billing/get_invoices.rb
PierrickVoulet: It is hard-coded.

examples/billing/get_invoices.rb
xflesym: Added comments.

examples/billing/get_invoices.rb
xflesym: Yup good call. I've made changes to use heredoc format instead.

mcloonan (approved):

PierrickVoulet (approved):

PR #277, opened 2020-12-11
Switch example CLI arguments to boolean (#265)

### LINKED ISSUE #265: True/false command line arguments are awkward in examples
See: https://github.com/googleads/google-ads-ruby/blob/master/examples/shopping_ads/add_shopping_product_ad.rb#L219

To leave it false, you have to not pass the command line that argument at all. To set it to true, you have to set the flag *and* a value, and the value doesn't matter. Even specifying "false" will result in the flag being set to true.

We should make this behave closer to boolean flags on other command line tools, where the act of setting the flag itself is enough to change it to true, with no value required.

abraham:
Closes #265 

```console
ruby examples/shopping_ads/add_shopping_product_ad.rb --create-default-listing-group true
ruby examples/shopping_ads/add_shopping_product_ad.rb --create-default-listing-group false

ruby examples/shopping_ads/add_shopping_product_listing_group_tree.rb --replace-existing-tree true
ruby examples/shopping_ads/add_shopping_product_listing_group_tree.rb --replace-existing-tree false

ruby examples/shopping_ads/add_shopping_smart_ad.rb  --create-default-listing-group true
ruby examples/shopping_ads/add_shopping_smart_ad.rb  --create-default-listing-group  false

### DIFF
--- a/examples/shopping_ads/add_shopping_smart_ad.rb
+++ b/examples/shopping_ads/add_shopping_smart_ad.rb
@@ -241,8 +241,8 @@ if __FILE__ == $0
     end
 
     opts.on('-d', '--create-default-listing-group CREATE-DEFAULT-LISTING-GROUP',
-        String, 'Create Default Listing Group') do |v|
-      options[:create_default_listing_group] = true if v
+        TrueClass, 'Create Default Listing Group') do |v|
+      options[:create_default_listing_group] = v
     end
 
     opts.separator ''
--- a/examples/shopping_ads/add_shopping_product_ad.rb
+++ b/examples/shopping_ads/add_shopping_product_ad.rb
@@ -215,8 +215,8 @@ if __FILE__ == $0
     end
 
     opts.on('-d', '--create-default-listing-group CREATE-DEFAULT-LISTING-GROUP',
-        String, 'Create Default Listing Group') do |v|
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: examples/shopping_ads/add_shopping_smart_ad.rb [hunk 1]
```diff
@@ -240,9 +240,8 @@ def add_default_shopping_listing_group(client, customer_id, ad_group_name)
       options[:merchant_center_id] = v
     end
 
-    opts.on('-d', '--create-default-listing-group CREATE-DEFAULT-LISTING-GROUP',
-        String, 'Create Default Listing Group') do |v|
-      options[:create_default_listing_group] = true if v
+    opts.on('-d', '--[no-]create-default-listing-group', 'Create Default Listing Group') do |v|
```
mcloonan, line 243: We have to conform to the exact format used in other languages, so that the functional testing framework doesn't have to change its implementation just for the Ruby case. Unfortunately, even though this is fairly elegant, we have to stick to a single `--create-default-listing-group` arg which then accepts a true/false value.

abraham: Updated

mcloonan (approved):

PR #285, opened 2021-02-02
Add region tags to get_invoices code example

PierrickVoulet:


examples/billing/get_invoices.rb
```diff
@@ -90,6 +93,7 @@ def get_invoices(customer_id, billing_setup_id)
     end
   end
 end
+# [END get_invoices_1]
```
mcloonan, line 96: Move this up 1 line so it matches the start of the block.

mcloonan (approved):

PR #287, opened 2021-02-18
version bump for v6.1

xflesym:


lib/google/ads/google_ads/version.rb
```diff
@@ -19,7 +19,7 @@
 module Google
   module Ads
     module GoogleAds
-      CLIENT_LIB_VERSION = '8.1.0'.freeze
+      CLIENT_LIB_VERSION = '8.2.0'.freeze
```
mcloonan, line 22: Make it 9.0.0 since we're removing v3 support.

mcloonan (changes requested):

lib/google/ads/google_ads/version.rb
xflesym: Done.

mcloonan (approved):

PR #302, opened 2021-04-29
add request id for non-streaming requests

### LINKED ISSUE #284: Request ID should be logged in log summary
It doesn't look like Request ID is in the log summary for non-streaming requests. We should always log request ID if we log anything.

xflesym:
Fixes #284

### DIFF
--- a/lib/google/ads/google_ads/interceptors/logging_interceptor.rb
+++ b/lib/google/ads/google_ads/interceptors/logging_interceptor.rb
@@ -312,11 +312,20 @@ module Google
                                 "no"
                               end
 
+            trailing_metadata = call
+              .instance_variable_get(:@wrapped)
+              .instance_variable_get(:@call)
+              .trailing_metadata
+
+            request_id = "N/A"
+            request_id = trailing_metadata["request-id"] if trailing_metadata
+
             [
               "CID: #{customer_id}",
               "Host: #{call.instance_variable_get('@wrapped').peer}",
               "Method: #{method}",
               "IsFault: #{is_fault_string}",
+              "Request ID: #{request_id}",
             ].join(", ")
           end
 


lib/google/ads/google_ads/interceptors/logging_interceptor.rb
```diff
@@ -58,6 +58,13 @@ def request_response(request:, call:, method:, metadata: {})
               if response.respond_to?(:partial_failure_error) && response.partial_failure_error
                 @logger.debug { build_partial_failure_message(response) }
               end
+              @logger.debug {
```
mcloonan, line 61: I think this should actually be built into the "build_summary_message" call above. I had to do some extra stuff for streaming requests because the request ID isn't available until the request is over, but for non-streaming requests I think the request ID is available as soon as we get the response.

mcloonan (changes requested):

lib/google/ads/google_ads/interceptors/logging_interceptor.rb
xflesym: Moved the logic to inside build_summary_message

lib/google/ads/google_ads/interceptors/logging_interceptor.rb
```diff
@@ -319,11 +312,20 @@ def build_summary_message(request, call, method, is_fault)
...
+              .instance_variable_get(:@wrapped)
+              .instance_variable_get(:@call)
+              .trailing_metadata
+
+            request_id = "N/A"
+            request_id = trailing_metadata["request-id"] if trailing_metadata
+
             [
               "CID: #{customer_id}",
               "Host: #{call.instance_variable_get('@wrapped').peer}",
```
mcloonan, line 325: Since we're using instance_variable_get('@wrapped') twice now, maybe we could extract this out above like you do with the request_id now.

mcloonan (changes requested):

lib/google/ads/google_ads/interceptors/logging_interceptor.rb
xflesym: Sure. I've extracted that out.

xflesym:
Mike - Could you take another look? I've addressed the change requested. Thanks

mcloonan (approved):

PR #304, opened 2021-05-06
Support Google Ads API v7.0

abraham:


ChangeLog
```diff
@@ -1,3 +1,6 @@
+11.0.0:
+ - Removed support for v4.
```
mcloonan, line 2: Make sure you update this in a followup commit to be more complete. It also looks like you haven't modified version.rb yet. These can both be done in a followup PR after this is merged.

mcloonan (approved):

PR #330, opened 2021-07-02
Changelog and version update

xflesym:


ChangeLog
```diff
@@ -1,3 +1,9 @@
+13.0.0:
+ - Removed support for v5.
+ - Updated get_ad_group_bid_modifier example to limit number of results in query.
```
mcloonan, line 3: We don't need to mention any of this in the changelog. The examples aren't bundled into the gem.

mcloonan: To be clear I meant the last 3 lines of the entry. I had them highlighted but I forgot that GitHub doesn't work that way :)

xflesym: Yup makes sense :) I've removed the 3 lines about examples.

mcloonan (approved):

krit8669:
What is all this bc I don't get on any of it or do any of this shit my
phone hacked and all u helping

krit8669:
I'm not subscribe to nothing my hacker is I aint

PR #333, opened 2021-07-14
Fix name collisions

### LINKED ISSUE #316: Add disambiguation mechanism to avoid resource name collision when using `client.resource.[type]`
The Ruby client library currently has a shortcut util for accessing resource types by `client.resource.[type]`.

However, when we have more than one resources sharing the same name, such as BiddingStrategy::TargetSpend and AccessibleBiddingStrategy::TargetSpend, the util syntax will cause ambiguity.

i.e. We are not able to specify the TargetSpend resource we need by using the syntax `client.resource.target_spend`
(Right now the workaround is to specify the full resource path, e.g. `Google::Ads::GoogleAds::V8::AccessibleBiddingStrategy::TargetSpend`

Please look into how to fix the factory util so that we can specify the exact type we need.

mcloonan:
Fixes #316

### DIFF
--- a/codegen/main.rb
+++ b/codegen/main.rb
@@ -60,9 +60,11 @@ Google::Ads::GoogleAds::KNOWN_API_VERSIONS.each do |version|
       require fn.gsub("#{GEM_ROOT}/lib/", "")
     end
   end
+
   resources = filter_resources_for_google_ads(version, potential_resources)
   resources = cleanup_paths(resources, :RESOURCE)
   resources, operations = filter_resources_into_resources_and_operations(resources)
+  resources = deduplicate_resources(resources)
 
   enums = filter_enums_for_google_ads(version, potential_enums)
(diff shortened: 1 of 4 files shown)


xflesym:
With this fix, what would be the actual usage for things like Google::Ads::GoogleAds::V8::Common::TargetCpa?
Is it `client.v8.common.target_cpa` ?

mcloonan:
Most things will be unchanged. The one in common will still be just `client.resource.target_cpa`, but the AccessibleBiddingStrategy one will be `client.resource.accessible_bidding_strategy_target_cpa`.

The logic is: if two leafs have the same name (in this case "target_cpa"), then replace each one with the full path from the "common" or "resource", in underscore case. If there's no name collision for an entry then nothing changes.

xflesym (approved):

PR #334, opened 2021-07-14
Path lookup automation

mcloonan:


mcloonan:
This may break people who were using kwargs for path lookups, but I think it's worth it. The kwarg names may change (such as campaign => campaign_id) in some cases.

xflesym (approved):

PR #356, opened 2021-11-04
Fix support for specifying an authentication hash for a service account

multiplegeorges:
It looks like #317 was incomplete.

The documentation says that we can specify the keyfile contents as a hash on `config.authentication`.

This PR fixes a bug so that works.


mcloonan:
The point of "authenticaton" is that you can do whatever logic you want in your config file (since it's a ruby file that is executed at runtime), and it passes the result of that directly through, untouched, to the underling gapic layer. So if you want to construct your own updater proc and not rely on ours, you're free to do so, but you should do that in the config file and not modify the library.

PR #368, opened 2021-12-21
add PMax retail example

xflesym:


mcloonan (approved):

abraham (approved):
There are couple of locations with mixed indentions (for example `create_and_link_image_asset`) an `_get_customer_conversion_goals` leads with an underscore while none of the other methods do.

It's cosmetic though so I'll leave it up to you if you want to fix it. 

It might a good idea to apply prettier or rubocop to the examples to create a more consistent experience.

PR #383, opened 2022-03-24
A few example updates.

mcloonan:
Change-Id: If5d88c9f8efa35a407a197190a000604189fa74c


AnashOommen (approved):

friedenberg (approved):
one question about constants vs symbols, but otherwise lgtm

PR #384, opened 2022-03-24
Adding example to demonstrate reusing page tokens.

mcloonan:
Change-Id: I76d851ffd94a7095fb6be90dff37af5567873ae5

### DIFF
--- a/examples/misc/navigate_search_result_pages_caching_tokens.rb
+++ b/examples/misc/navigate_search_result_pages_caching_tokens.rb
@@ -36,10 +36,10 @@ def navigate_search_result_pages_caching_tokens(customer_id)
   # ENV['HOME']/google_ads_config.rb when called without parameters
   client = Google::Ads::GoogleAds::GoogleAdsClient.new
 
-  # The cache of page tokens. The key is the 0-indexed page number.
+  # The cache of page tokens. The key is the page number.
   # The first page's token is always the empty string.
   page_tokens = {
-    0 => ''
+    1 => ''
   }
 
   puts "Fetching page 1 to get metadata"
@@ -51,7 +51,7 @@ def navigate_search_result_pages_caching_tokens(customer_id)
     return_total_results_count: true,
   )
 
-  cache_next_page_token(page_tokens, response.page, 1)
+  cache_next_page_token(page_tokens, response.page, 2)
 
   # The total results count does not take into consideration the LIMIT clause
   # of the query so we need to find the minimal value between the limit and the
@@ -94,7 +94,7 @@ end
 # [START navigate_search_result_pages_caching_tokens]
 def fetch_and_print_page_results(client, customer_id, query, page_size,
                                  return_total_results_count, page_number, page_tokens)
-  if page_tokens.has_key?(page_number - 1)
+  if page_tokens.has_key?(page_number)
     puts 'The page token for the request page was cached. Reusing it.'
     current_page = page_number
   else
@@ -104,15 +104,15 @@ def fetch_and_print_page_results(client, customer_id, query, page_size,
   end
 
   while current_page <= page_number
-    puts "Fetching page #{current_page}."
+    puts page_tokens
     response = client.service.google_ads.search(
       customer_id: customer_id,
       query: query,
       page_size: page_size,
       return_total_results_count: return_total_results_count,
-      page_token: page_tokens[current_page - 1],
+      page_token: page_tokens[current_page],
     )
-    cache_next_page_token(page_tokens, response.page, current_page)
+    cache_next_page_token(page_tokens, response.page, current_page + 1)
     current_page += 1
   end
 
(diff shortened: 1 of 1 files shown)


examples/misc/navigate_search_result_pages_caching_tokens.rb
```diff
@@ -0,0 +1,186 @@
...
+      page_number,
+      page_tokens,
+    )
+  end
+end
+
+# [START navigate_search_result_pages_caching_tokens]
+def fetch_and_print_page_results(client, customer_id, query, page_size,
+                                 return_total_results_count, page_number, page_tokens)
+  if page_tokens.has_key?(page_number)
```
PierrickVoulet, line 97: Should not `page_number` be `page_number  - 1`?

PierrickVoulet (changes requested):

examples/misc/navigate_search_result_pages_caching_tokens.rb
mcloonan: Good catch, thanks! It *looked* like it was working with this mistake but there was a subtle error.

PierrickVoulet (approved):

examples/misc/navigate_search_result_pages_caching_tokens.rb
```diff
@@ -0,0 +1,186 @@
...
+    customer_id,
+    query,
+    PAGE_SIZE,
+    true,
+    middle_page,
+    page_tokens,
+  )
+
+  puts 'Print all pages, starting at the last page and moving to the first.'
+  (1..total_pages).reverse_each do |page_number|
```
jradcliff, line 81: Should that `1` be a `0`? Otherwise you won't print the first page.

examples/misc/navigate_search_result_pages_caching_tokens.rb
```diff
@@ -0,0 +1,186 @@
...
+  puts "Fetching page 1 to get metadata"
+  query = "SELECT campaign.id, campaign.name FROM campaign ORDER BY campaign.name LIMIT #{RESULTS_LIMIT}"
+  response = client.service.google_ads.search(
+    customer_id: customer_id,
+    query: query,
+    page_size: PAGE_SIZE,
+    return_total_results_count: true,
+  )
+
+  cache_next_page_token(page_tokens, response.page, 1)
```
jradcliff, line 54: Should the argument here be `0`, since you're adding `1` in `cache_next_page_token()`?

examples/misc/navigate_search_result_pages_caching_tokens.rb
mcloonan: "page_tokens" is 0-indexed, cache_next_page_token expects a 0-indexed page number, and fetch_and_print_page_results expects a 1-indexed page number.

It all works the way it's written, but it is a little confusing. This is following the design in the golden.

examples/misc/navigate_search_result_pages_caching_tokens.rb
mcloonan: No, 1 is correct because this is for page 2 and cache_next_page_token is 0-indexed. More details on other comment.

examples/misc/navigate_search_result_pages_caching_tokens.rb
jradcliff: That seems awfully confusing. If you think Ruby users would be less confused if you made `page_tokens` and its usages 1-indexed, please go ahead and make that change. I suspect the PHP golden did it this way because it was using a 0-based array instead of a map (or chose to use 0-based keys in a map), but I don't think you have to follow that exactly.
@PierrickVoulet , please chime in if I missed something.

mcloonan: That's actually a good point. My programmer wanted everything to be 0-indexed but the 1-indexed approach makes more sense here. I updated it.

PierrickVoulet: Java implemented the 1-indexed solution, feel free to replicate https://github.com/googleads/google-ads-java/blob/main/google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java#L109.

I agree this can be confusing, I will make the improvement in PHP as well.

PierrickVoulet: Here is how the change looks like in PHP: https://github.com/googleads/google-ads-php/pull/777.

jradcliff (approved):

examples/misc/navigate_search_result_pages_caching_tokens.rb
jradcliff: Nice - much easier to interpret now. Thank you both!

AnashOommen (approved):

PR #393, opened 2022-05-04
Updating authentication example to work with json files, and cover bo…

mcloonan:
…th web and installed flows.

Change-Id: I8b1de35a0961c4aedf2c7090fdd705c709e2224c


examples/authentication/generate_user_credentials.rb
```diff
@@ -27,16 +27,31 @@
...
 
-def authenticate_in_web_application(client_id, client_secret, port)
-  callback_uri = sprintf('http://localhost:%s', port)
+def generate_user_credential(path)
+  json = JSON.load_file(path)
+  creds, callback_uri, port = if !json["installed"].nil?
+            [json["installed"], "http://#{SERVER}:#{PORT}", PORT]
+          elsif !json["web"].nil?
+            web_creds = json["web"]
+            uri = json["web"]["redirect_uris"].first
```
jradcliff, line 40: There's a slight risk that you'll pick the wrong `uri` here if the client has multiple redirect URIs. Since this example relies on the callback to SERVER and PORT, WDYT of just using that for the `uri` in both cases?

mcloonan: I actually use the URI and Port extracted from here down below (and tested this), so that's not a concern.

Maybe a better solution is to add a comment mentioning that if there's more than one redirect URI, then they should add additional logic here to choose the correct one?

AnashOommen (dismissed):

examples/authentication/generate_user_credentials.rb
jradcliff: The comment approach LGTM. Can always make this fancier with a URL command line param later if you find users are frequently using OAuth clients with multiple redirect URIs.

jradcliff (approved):

PR #408, opened 2022-09-12
Relaxing dependency on gapic-common so our library is compatible with…

### LINKED ISSUE #403: Do not restrict minor version of `gapic-common` to 0.6.x
**Describe the bug:**

This gem has a strict dependency of the `gapic-common` gem on the minor version 0.6.x, i.e. https://github.com/googleads/google-ads-ruby/blob/afa41190a2c0646f511221a1746f365b90173649/google-ads-googleads.gemspec#L37. I don't see why it would be necessary to restrict the minor version of this gem. This is preventing us from updating other gems that depends on a higher minor version than 0.6 of the `gapic-common` gem. For example, the latest release of the [google-analytics-admin-v1alpha](https://github.com/googleapis/google-cloud-ruby/tree/main/google-analytics-admin-v1alpha) gem has the following dependency `"gapic-common", ">= 0.10", "< 2.a"`. The strict dependency of this gem prevents us from using the latest version of the `google-analytics-admin-v1alpha` gem.

**Steps to Reproduce:**

...

**Expected behavior:**

I would expect there to only be a dependency on the major version, i.e. `'gapic-common', '~> 0.6'`

mcloonan:
… more other libraries.

Fixes #403 

Change-Id: Ic8c711f5f3fc6fb617ef23363792708ac24e0ba9

### DIFF
--- a/google-ads-googleads.gemspec
+++ b/google-ads-googleads.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.5.0'
   s.summary = 'Google client library for the Google Ads API'
 
-  s.add_dependency 'gapic-common', '~> 0.6.0'
+  s.add_dependency 'gapic-common', '~> 0.6'
   s.add_dependency 'google-protobuf', ['>= 3.19.4', '< 4.0']
 
   s.add_development_dependency 'bundler', ["> 1.9", "< 3"]


jradcliff (approved):

alexibz:
This is great, is there any timeline for when the new version of the gem with this change would be released?

mcloonan:
I just released a new version, 19.0.0, now, which includes this change.

PR #420, opened 2022-11-14
Fixing an issue with the field mask util and updating to 'false'.

mcloonan:
Change-Id: I861dc5dc7651e4d9555d65bbfa167b6e63d06967


mcloonan:
Turns out this would not work in some edge cases. I'll just update the example to show best practices instead.

PR #430, opened 2023-03-01
Use the experiment arm response to get the draft campaign ID, instead…

mcloonan:
… of a followup search query.

Change-Id: I338c5b265a96045f231ce3b67465d524244afe69

### DIFF
--- a/examples/campaign_management/create_experiment.rb
+++ b/examples/campaign_management/create_experiment.rb
@@ -27,8 +27,7 @@ def create_experiment(customer_id, base_campaign_id)
   client = Google::Ads::GoogleAds::GoogleAdsClient.new
 
   experiment = create_experiment_resource(client, customer_id)
-  treatment_arm = create_experiment_arms(client, customer_id, base_campaign_id, experiment)
-  draft_campaign = fetch_draft_campaign(client, customer_id, treatment_arm)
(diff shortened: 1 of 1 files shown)


examples/campaign_management/create_experiment.rb
```diff
@@ -86,47 +85,24 @@ def create_experiment_arms(client, customer_id, base_campaign_id, experiment)
...
+  control_arm_result = response.results.first
+  treatment_arm_result = response.results.last
 
-  puts "Created control arm with resource name #{control_arm}."
-  puts "Created treatment arm with resource name #{treatment_arm}."
+  puts "Created control arm with resource name #{control_arm_result.resource_name}."
+  puts "Created treatment arm with resource name #{treatment_arm_result.resource_name}."
 
-  treatment_arm
+  treatment_arm_result.experiment_arm.in_design_campaigns.first
```
jradcliff, line 102: Is the `in_design_campaigns` attribute mutable? If not, it won't be returned even with `response_content_type = MUTABLE_RESOURCE`.

mcloonan: It is not mutable, and it is returned anyway for this feature specifically for this specific use case (to avoid a followup request which would need to be done pretty much every time you ever make an experiment arm).

PR #433, opened 2023-03-14
Add PMax for Travel Goals example

danielfrg:
Output

CleanShot 2023-03-14 at 12 55 34@2x

CleanShot 2023-03-14 at 12 56 02@2x

### DIFF
--- a/examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
+++ b/examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
@@ -742,26 +742,14 @@ end
 
 if __FILE__ == $0
   options = {}
-  # The following parameter(s) should be provided to run the example. You can
-  # either specify these by changing the INSERT_XXX_ID_HERE values below, or on
-  # the command line.
-  #
-  # Parameters passed on the command line will override any parameters set in
-  # code.
-  #
-  # Running the example with -h will print the command line usage.
-  options[:customer_id] = 'INSERT_CUSTOMER_ID_HERE'
-
-  # Sets a place ID that uniquely identifies a place in the Google Places database.
-  # See https://developers.google.com/places/web-service/place-id to learn more.
-  # The provided place ID must belong to a hotel property.
-  options[:place_id] = 'INSERT_PLACE_ID_HERE'
 
   OptionParser
     .new do |opts|
       opts.banner = format('Usage: %s [options]', File.basename(__FILE__))
+
       opts.separator ''
       opts.separator 'Options:'
+
       opts.on('-C', '--customer-id CUSTOMER-ID', String, 'Customer ID') do |v|
         options[:customer_id] = v
       end
--- a/examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
+++ b/examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 # Encoding: utf-8
 #
-# Copyright 2018 Google LLC
+# Copyright 2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -197,7 +197,6 @@ def create_hotel_asset_set(client, customer_id)
     )
 
   # Prints some information about the response.
-  print_response_details(response)
   response.results.first.resource_name
 end
 # [END create_hotel_asset_set]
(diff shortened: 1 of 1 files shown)


examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
```diff
@@ -0,0 +1,823 @@
+#!/usr/bin/env ruby
+# Encoding: utf-8
+#
+# Copyright 2018 Google LLC
```
mcloonan, line 4: 2023

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
```diff
@@ -0,0 +1,823 @@
...
+# [START get_hotel_asset_suggestion]
+def get_hotel_asset_suggestion(client, customer_id, place_id)
+  response =
+    client.service.travel_asset_suggestion.suggest_travel_assets(
+      customer_id: customer_id,
+      language_option: 'en-US',
+      place_id: [place_id]
+    )
+
+  response.hotel_asset_suggestions[0]
```
mcloonan, line 341: We usually use `.first` in ruby examples for this.

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
```diff
@@ -0,0 +1,823 @@
...
+  if hotel_asset_suggestion.status == :SUCCESS
+    hotel_asset_suggestion.text_assets.each do |hotel_text_asset|
+      if hotel_text_asset.asset_field_type == :HEADLINE ||
+         hotel_text_asset.asset_field_type == :DESCRIPTION
+        # Headlines and descriptions were already created at the first step of
+        # this code example.
+        next
+      end
+
+      puts format(
```
mcloonan, line 543: Prefer in-line replacement.

`"A text asset with text '#{hotel_text_asset.text}' is ..."`

mcloonan (commented):
Only a few minor comments. Thanks for writing this!

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
danielfrg: done

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
danielfrg: done

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
```diff
@@ -0,0 +1,804 @@
...
+  options = {}
+  # The following parameter(s) should be provided to run the example. You can
+  # either specify these by changing the INSERT_XXX_ID_HERE values below, or on
+  # the command line.
+  #
+  # Parameters passed on the command line will override any parameters set in
+  # code.
+  #
+  # Running the example with -h will print the command line usage.
+  options[:customer_id] = 'INSERT_CUSTOMER_ID_HERE'
```
mcloonan, line 753: Leave these placeholders out so command line input is the only way, and also remove them if you're modifying an example that still has them in the future. We decided to remove this going forward but not to spend the time explicitly removing it if we're not touching the example for some other reason.

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
```diff
@@ -0,0 +1,804 @@
...
+
+# Prints the details of a MutateGoogleAdsResponse.
+def print_response_details(response)
+  # Parse the mutate response to print details about the entities that
+  # were created by the request.
+  suffix = '_result'
+  response.mutate_operation_responses.each do |result|
+    result
+      .to_h
+      .select { |_k, v| v }
```
mcloonan, line 729: I think this should be map instead of select, since there is no case where you're removing an element.

mcloonan (approved):

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
danielfrg: I copied this one from another example but it does filter for the case where `v` is empty.
Without that it prints a lot more of created resources.

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
danielfrg: Makes sense. I copied that section. I will update others as/if go through them.

examples/hotel_ads/add_performance_max_for_travel_goals_campaign.rb
mcloonan: Good point, I didn't consider the case that v would be empty. LGTM then.

mcloonan (approved):

PR #459, opened 2023-10-20
Update V15 example ProductCategoryConstant

danielfrg:
Renaming to ProductCategoryConstant for V15.

### DIFF
--- a/examples/shopping_ads/get_product_category_constant.rb
+++ b/examples/shopping_ads/get_product_category_constant.rb
@@ -63,16 +63,12 @@ def get_product_category_constant(customer_id)
   response.each do |row|
     product_category = row.product_category_constant
 
-    localized_name = ""
-    product_category.localizations.each do |l|
-      if l.region_code == "US" && l.language_code == 'en'
-        localized_name = l.value
-        break
-      end
-    end
+    localized_product = product_category.localizations.detect{ |l|
+      l.region_code == "US" && l.language_code == 'en'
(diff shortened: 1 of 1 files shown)


examples/shopping_ads/get_product_category_constant.rb
```diff
@@ -63,17 +61,17 @@ def get_product_bidding_category_constant(customer_id)
   root_categories = Set.new
 
   response.each do |row|
-    product_bidding_category = row.product_bidding_category_constant
+    product_category = row.product_category_constant
     category = {
-      name: product_bidding_category.localized_name,
-      id: product_bidding_category.resource_name,
+      name: product_category.localizations,
```
jradcliff, line 66: The `localizations` field is a repeated field, so you need to find the `en`/`US` entry in that collection

danielfrg: Done

jradcliff: Would `select` followed by `first` be more idiomatic for Ruby?

danielfrg: I usually try to keep those a bit more generic but I changed it to `detect` which i think is more idiomatic.

jradcliff: Thanks! The updated version looks easier to read.

jradcliff (approved):

PR #463, opened 2023-10-20
V15: Update examples to add DMA consent field

danielfrg:


examples/remarketing/upload_call_conversion.rb
```diff
@@ -149,18 +159,24 @@ def upload_call_conversion(
       options[:conversion_value] = v.to_f
     end
 
-    opts.on('-d', '--conversion-custom-variable-id CONVERSION-CUSTOM-VARIABLE-ID', \
+    opts.on('-w', '--conversion-custom-variable-id CONVERSION-CUSTOM-VARIABLE-ID', \
```
danielfrg, line 162: Changed this to match what we do in Python

examples/remarketing/upload_store_sales_transactions.rb
```diff
@@ -267,6 +276,19 @@ def build_offline_user_data_job_operations(
       t.transaction_date_time = "2020-05-01 23:52:12"
       t.custom_value = custom_value unless custom_value.nil?
     end
+    if !ad_user_data_consent.nil? || !ad_personalization_consent.nil?
+      job.consent = client.resource.consent do |c|
```
jradcliff, line 280: Could you rename the `job` variable to something more intuitive, like `op` or `operation`? I realize the name was being used already, but it makes it confusing for users because the block is operating on a single operation in the job, not the job itself.

danielfrg: Agreed, i only kept it because it was already there but agree it was confusing. It's done now.

jradcliff (approved):

examples/remarketing/upload_store_sales_transactions.rb
jradcliff: Thanks!

PR #467, opened 2023-11-14
Support direct access with Service Account

danielfrg:
Removes a check we for the `impersonate` field in order to make it work as expected: By only settings the `keyfile` field.

### DIFF
--- a/lib/google/ads/google_ads/google_ads_client.rb
+++ b/lib/google/ads/google_ads/google_ads_client.rb
@@ -225,7 +225,6 @@ module Google
         # Provides a Google::Auth::Credentials initialized with a keyfile
         # specified in the config.
         def get_service_account_credentials
-          raise 'config.impersonate required if keyfile specified' unless @config.impersonate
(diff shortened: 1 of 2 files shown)


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -225,7 +225,6 @@ def get_updater_proc
         # Provides a Google::Auth::Credentials initialized with a keyfile
         # specified in the config.
         def get_service_account_credentials
-          raise 'config.impersonate required if keyfile specified' unless @config.impersonate
```
jradcliff, line 228: Could you add a test to verify this works as expected?

https://github.com/googleads/google-ads-ruby/blob/main/test/test_google_ads_client.rb

danielfrg: Updated the test. It still will raise for an OpenSSL error but that means its actually reading the test file so the `keyfile` config is valid.

jradcliff (approved):

PR #468, opened 2023-11-14
Support for use_cloud_org_for_api_access

danielfrg:


lib/google/ads/google_ads/google_ads_client.rb
```diff
@@ -103,6 +103,7 @@ def load_environment_config
           @config.developer_token = ENV.fetch("GOOGLE_ADS_DEVELOPER_TOKEN", @config.developer_token)
           @config.login_customer_id = ENV.fetch("GOOGLE_ADS_LOGIN_CUSTOMER_ID", @config.login_customer_id)
           @config.linked_customer_id = ENV.fetch("GOOGLE_ADS_LINKED_CUSTOMER_ID", @config.linked_customer_id)
+          @config.use_cloud_org_for_api_access = ENV.fetch("GOOGLE_ADS_USE_CLOUD_ORG_FOR_API_ACCESS", @config.use_cloud_org_for_api_access)
```
jradcliff, line 106: @AnashOommen , do we want to support setting this via an environment variable? I don't see that in the golden: http://tg/1985917.

AnashOommen: For now, let's keep only minimal support for this setting.

danielfrg: Sounds good. Removed the env var option for this var.

jradcliff (approved):

PR #486, opened 2024-03-21
Update google-ads-googleads.gemspec

himanshu-patel-dev:
Updating `gapic-common` right version from `0.20` to `0.21.1`

### DIFF
--- a/google-ads-googleads.gemspec
+++ b/google-ads-googleads.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.5.0'
   s.summary = 'Google client library for the Google Ads API'
 
-  s.add_dependency 'gapic-common', '~> 0.20'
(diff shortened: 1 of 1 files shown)


google-ads-googleads.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.5.0'
   s.summary = 'Google client library for the Google Ads API'
 
-  s.add_dependency 'gapic-common', '~> 0.20'
+  s.add_dependency 'gapic-common', '~> 0.21.1'
```
mcloonan, line 37: change it to `['>= 0.21.1', '< 1.0']`

The `~>` operator isn't going to do what we want here, and will be overly restrictive

himanshu-patel-dev: I addressed the comment. but I have 2 identical commit messages. Should I squash them?

mcloonan:
No worries about the commit messages, and thanks for helping out here!

PR #509, opened 2025-03-04
fix incorrect setting of the email field

k0va1:


dorasun (approved):
Thanks for catching this and making the changes!

PR #514, opened 2025-04-25
Update example

dorasun:
Update upload ECL example to include session attribute inputs

### DIFF
--- a/examples/remarketing/upload_enhanced_conversions_for_leads.rb
+++ b/examples/remarketing/upload_enhanced_conversions_for_leads.rb
@@ -117,6 +117,7 @@ def upload_conversion_with_identifiers(
       cc.class.module_eval { attr_accessor :session_attributes_key_value_pairs}
       cc.session_attributes_key_value_pairs = ::Google::Ads::GoogleAds::V19::Services::SessionAttributesKeyValuePairs.new
       
+      # Loop thru inputted session_attributes_hash to populate session_attributes_key_value_pairs
       session_attributes_hash.each do |key, value|
         pair = ::Google::Ads::GoogleAds::V19::Services::SessionAttributeKeyValuePair.new
         pair.session_attribute_key = key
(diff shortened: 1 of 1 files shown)


examples/remarketing/upload_enhanced_conversions_for_leads.rb
```diff
@@ -96,6 +105,27 @@ def upload_conversion_with_identifiers(
...
+    # [START add_session_attributes]
+    # Set one of the session_attributes_encoded or
+    # session_attributes_key_value_pairs fields if either are provided.
+    if session_attributes_encoded != nil
+      cc.class.module_eval { attr_accessor :session_attributes_encoded}
+      cc.session_attributes_encoded = session_attributes_encoded
+    elsif session_attributes_hash != nil
+      # Add new attribute to click conversion object
+      cc.class.module_eval { attr_accessor :session_attributes_key_value_pairs}
+      cc.session_attributes_key_value_pairs = ::Google::Ads::GoogleAds::V19::Services::SessionAttributesKeyValuePairs.new
```
mcloonan, line 118: This seems to be the best way to do this right now, but you should look into upgrading the factory generator to capture this and provide a shortcut like we do for most other things like client.service.[whatever], client.resource.[whatever], and client.operation.[whatever]

This should probably be clumped in with "resource".

examples/remarketing/upload_enhanced_conversions_for_leads.rb
```diff
@@ -226,6 +268,15 @@ def normalize_and_hash_email(email)
       exit
     end
   end.parse!
+    
+  if (options[:session_attributes_encoded]!=nil && options[:session_attributes_hash]!=nil)
```
mcloonan, line 272: I think it's more canonical to do:

!options[:session_attributes_encoded].nil?

mcloonan (changes requested):

examples/remarketing/upload_enhanced_conversions_for_leads.rb
dorasun: Thanks for the suggestion, I'll take a look at doing that in the future

examples/remarketing/upload_enhanced_conversions_for_leads.rb
dorasun: Thanks for that suggestion, updated!

mcloonan (approved):

Issue #502, opened 2025-01-09
Update gem to be dependent of google-protobuf > 4.0


nick-lawrence-ctm:
**Describe the problem you are trying to solve:**
Trying to install gem on amazonlinux 2023 using ruby 3.4.1 and cannot due to compatibility issues with installing google-protobuf 3.25.5

**Describe the solution you'd like:**
We would like to be able to install the latest google-protobuf (4.29.3) to get around to ruby 3.4.1 incompatability, but this gem's dependencies prevent us from doing so.

mcloonan:
Thanks for pointing this out. We'll take as soon as we can.

mcloonan:
We're still looking into it and resolving some other dependency issues with the change.

njaremko:
Pinging this to see if there's been any progress on this, since it's been a month

Vagab:
hey @mcloonan, is there anything that we can help here with maybe? Would love to upgrade to ruby 3.4

larouxn:
👋 Hey folks, I opened the following pull request. Unsure if more needs to be done re: https://github.com/googleads/google-ads-ruby/issues/502#issuecomment-2634332995. Would also appreciate some real world testing of said change with Ruby 3.4.x as I don't have an application for such these days.

- https://github.com/googleads/google-ads-ruby/pull/515

dorasun:
Thanks for creating a PR, I will try to take a look at it within this week. Just a heads up that it may take some time to fully test the changes to ensure there's no impact

dorasun:
Unfortunately I will likely not be able to take a look at the PRs until late May/early June. Sorry for the inconvenience!

jtarchie:
Does v34.0.0 sneak in the support we need?

larouxn:
Doesn't look like it based on the v33.1.0 vs v34.0.0 diff view and the lack of any change to the gemspec.

https://github.com/googleads/google-ads-ruby/compare/google-ads-googleads/v33.1.0...google-ads-googleads/v34.0.0

https://github.com/googleads/google-ads-ruby/blob/59dfe4d049c603e4ba323ce599ce04702b08b450/google-ads-googleads.gemspec#L37-L38

PR #515, opened 2025-05-09
Fix support for Ruby 3.4

larouxn:
**Problem**

`google-ads-ruby` cannot be installed when using Ruby 3.4.x. This is due to the locking of the upper bound version of `google-protobuf` dependency to `< 4.0.0` and `gapic-common` dependency to `< 1.0.0`. 

**Solution**

Bump the upper bound version lock of both of the aforementioned dependencies one major version, thus allowing for `>= 4.0.0` of `google-protobuf` and `>= 1.0.0` of `gapic-common`.

- Unlock gapic-common from < 1.0.0 to < 2.0.0.
- Unlock google-protobuf from < 4.0.0 to < 5.0.0.
  - google-protobuf 4.0.0 is required for Ruby 3.4 compatibility.

Resolves https://github.com/googleads/google-ads-ruby/issues/502

**Tophatting**

Before (Ruby 3.3.x installs fine, Ruby 3.4.x fails to install)

Screenshot From 2025-05-09 08-41-14
Screenshot From 2025-05-09 08-41-38
Screenshot From 2025-05-09 08-41-44

After (Ruby 3.4.x installs fine, using this branched version)

Screenshot From 2025-05-09 08-54-07

Testing - Simply ran the tests in this repo via `bundle exec rake test`.

Screenshot From 2025-05-09 08-59-32

Notes

- I no longer have a production application I can test this change against. If someone does have an application with a solid test suite they can run this change against or perhaps toss this on a staging application or otherwise, that would be very helpful.

- If you look closely during my testing above there's a small working directory diff. That's simply me commenting out the `allocation_tracker` gem as it no longer installs properly. Seems like such has been an issue for close to 9 months now: https://github.com/ko1/allocation_tracer/issues/19.

Screenshot From 2025-05-09 08-57-47

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ See the [Logging guide][].
 
 ## Ruby versions
 
-The client library supports versions 3.1, 3.2, and 3.3 of Ruby. We plan to
+The client library supports versions 3.2, 3.3, and 3.4 of Ruby. We plan to
 support three Ruby releases at any one time. As Ruby releases once a year on
 December 25th, we will look at dropping support for the oldest version early in
 the following year.
--- a/google-ads-googleads.gemspec
+++ b/google-ads-googleads.gemspec
@@ -34,8 +34,8 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 3.0.0'
   s.summary = 'Google client library for the Google Ads API'
 
-  s.add_dependency 'gapic-common', ['>= 0.25', '< 1.0']
-  s.add_dependency 'google-protobuf', ['>= 3.19.4', '< 4.0']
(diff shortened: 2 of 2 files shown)


dorasun:
Thanks for creating this PR. Unfortunately I may not have the bandwidth to fully review this until late May/early June, but I will get to it as soon as I can!

jtarchie:
I tried to this via a `Gemfile` via

```ruby
gem 'google-ads-googleads', require: "google/ads/google_ads", github: 'googleads/google-ads-ruby', ref: 'c724e3f17b92b6f5357a3e0d10112e50fa4c38fb'
```

It does not work because there is a code gen step missing when pulling from source. This is obvious once I realized there was one. Just commenting so other know, you cannot just add this PR and have it work.

PR #537, opened 2026-02-10
Update upload_enhanced_conversions_for_leads.rb

BenRKarl:
Add comment specifying that session attributes is only available by allowlist.


dorasun (approved):
LGTM!

